PR #52, opened 2022-01-11
Support for Yggdrasil network

cofob:


cofob:
Please, do not use it. New version here: https://git.sr.ht/~cofob/docker-images/tree/master/item/vpn/yggdrasil/yggdrasil-auto/Dockerfile

PR #56, opened 2022-02-19
Dev killswitch

pokamest:
KillSwitch feature


pokamest:
See https://github.com/amnezia-vpn/amnezia-client/pull/444

PR #66, opened 2022-04-12
fixed line endings for windows drivers *.inf files

wslcdg:
Explicitly set cr-lf line endings for windows.
Possible solution for https://github.com/amnezia-vpn/desktop-client/issues/64


pokamest:
Merged to dev

PR #90, opened 2022-08-10
AVP-01-001 (Fix permissions for Linux and MacOS)

leetthewire:


pokamest:
Qt Installer Framework is a reason of this issue. Latest QIF version set correct permissions to file - owner root)

PR #133, opened 2022-11-19
GitHub actions

pokamest:
GitHub actions

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -67,7 +67,7 @@ jobs:
   #   - name: Build project
   #     run: bash deploy/build_linux.sh
 
-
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,256 @@
...
+
+    - name: 'Setup Java'
+      uses: actions/setup-java@v3
+      with:
+        distribution: 'temurin'
+        java-version: '8'
+
+    - name: 'Build project'
+      run: |
+        export NDK_VERSION=21d
```
Nethius, line 245: we have to install an older version of the NDK because of this bug https://bugreports.qt.io/browse/QTBUG-104580

PR #157, opened 2023-01-20
Build openvpn android native library with pt transport from sources

outspace:
Remove prebuilded android native libraries and build them from sources. 
Change openvpn native libraries for libraries with pt transport.
Build openvpn pt transport plugin from sources.


kolobchanin:
I think there are some redudant files on that path: client/android/cpp/pluginjni/[arm64-v8a, armeabi-v7a, x86, x86_64]/libck-ovpn-plugin.so

PR #168, opened 2023-02-06
feature/port-availability-on-container-setup

Nethius:


client/core/servercontroller.cpp
```diff
@@ -856,3 +862,45 @@ QString ServerController::replaceVars(const QString &script, const Vars &vars)
...
+        break;
+    }
+    case DockerContainer::Cloak:{
+        script.append("':$CLOAK_SERVER_PORT'");
+        break;
+    }
+    default:
+        return ErrorCode::NoError;
+    }
+
```
pokamest, line 898: IKEv2, DNS, SFTP containers missing, also missing protocol TCP/UDP detection.
You need to use "port" and "transport_proto" that is already present in each container config.
For containers like IKEv2 where ports are fixed, and in amnezia's config it's set to 0, I suggest to implement some new function like ContainerProps::fixedPorts(DockerContainer c) which will return list of fixed ports.
When done, you will be able to get rid of this boilerplate, just add two checks - for actual container ports, and for fixed ports. I guess we will implement new containers with fixed ports sometime.

pokamest (changes requested):

client/containers/containers_defs.cpp
```diff
@@ -27,6 +27,16 @@ QString ContainerProps::containerToString(amnezia::DockerContainer c){
     return "amnezia-" + containerKey.toLower();
 }
 
+QString ContainerProps::containerTypeToString(amnezia::DockerContainer c){
```
pokamest, line 30: This function don't needed
You can use ContainerProps::defaultProtocol to get default Protocol for container, and ProtocolProps::protoToString if needed

pokamest (changes requested):

pokamest (approved):

PR #179, opened 2023-03-02
feature/linux-wireguard-implementation

Nethius:


ipc/ipcserver.cpp
```diff
@@ -124,3 +125,53 @@ void IpcServer::setLogsEnabled(bool enabled)
...
+    QProcess copyWireguardConfigProcess;
+
+    bool errorOccurred = false;
+
+    connect(&copyWireguardConfigProcess, &QProcess::errorOccurred, this, [&errorOccurred](QProcess::ProcessError error) {
+        qDebug() << "WireguardProtocol::WireguardProtocol error occured while copying wireguard config: " << error;
+        errorOccurred = true;
+    });
+
+    copyWireguardConfigProcess.setProgram("/bin/cp");
```
pokamest, line 141: Why not just use QFile::copy ?
What if dest file already exists?

pokamest (changes requested):

PR #186, opened 2023-03-18
Fixed incorrect formation of certificate registration utility startup…

JRaerty:
… string


pokamest:
Merged to dev instead

PR #187, opened 2023-03-18
Plain ShadowSocks support

outspace:
This PR is about plain ShadowSocks support.
Now we get shadowsocks to work correctly, but we still need to do some work on it. 

TODO:
R̶e̶m̶o̶v̶e̶ O̶p̶e̶n̶V̶P̶N̶ f̶r̶o̶m̶ d̶o̶c̶k̶e̶r̶ c̶o̶n̶t̶a̶i̶n̶e̶r̶. 
S̶e̶t̶ D̶N̶S̶ f̶r̶o̶m̶ a̶ l̶i̶s̶t̶ (̶n̶o̶w̶ i̶t̶ u̶s̶e̶s̶ o̶n̶l̶y̶ g̶o̶o̶g̶l̶e̶ D̶N̶S̶ (̶8̶.8̶.8̶.8̶)̶.
T̶e̶s̶t̶ i̶t̶. 
MacOS/iOS support


client/android/shadowsocks/src/main/java/org/amnezia/vpn/shadowsocks/core/Core.kt
```diff
@@ -161,7 +161,7 @@ object Core : Configuration.Provider {
                             NotificationManager.IMPORTANCE_LOW),
                     SubscriptionService.notificationChannel))
             notification.deleteNotificationChannel("service-nat")   // NAT mode is gone for good
-        }
+        } */
```
kolobchanin, line 164: Why you commented this part of code?

client/android/build.gradle
```diff
@@ -22,6 +22,8 @@ buildscript {
     }
 
     dependencies {
+    	classpath 'com.google.gms:google-services:4.3.2'
+    	classpath 'com.google.firebase:firebase-crashlytics-gradle:2.1.1'
```
kolobchanin, line 26: Maybe we should uncomment Firebase init, which was turned off in previos commit?

client/android/src/org/amnezia/vpn/VPNService.kt
```diff
@@ -802,6 +802,8 @@ Log.i(tag, "shadowsocksConfig: profile.remotePort" + profile.remotePort)
                     }
             if (!profile.bypass) builder.addAllowedApplication(me)
         }
+        val me = packageName
+        builder.addDisallowedApplication(me)
```
kolobchanin, line 806: This can be written in one line

client/android/shadowsocks/src/main/java/org/amnezia/vpn/shadowsocks/core/Core.kt
outspace: Because we don't use notification channels in our app and I want to save maximum of original SS code.

client/android/build.gradle
outspace: Crashlytics is a good thing, but I'm not sure that collecting of some logs is fit into our ideology. We can register new [google-services.json](https://github.com/amnezia-vpn/amnezia-client/pull/187/files#diff-06a86e87a0d834bd8d75c40dcadcbfb80c6415cd3faca61fd4db82d0c0e3bb37) into corporete acc and turn it on.

client/android/src/org/amnezia/vpn/VPNService.kt
outspace: yep, in general work with profile.bypass can be better.

PR #197, opened 2023-03-26
v2ray Trojan container

vysotskylev:


client/client.pro
Nethius, line 1: please use CMake as we won't be using QMake anymore

PR #224, opened 2023-04-18
Prevent ipv6 leak

outspace:
We can use block-ipv6 option only with precompiled OpenVPN,
because this option is supported from OpenVPN 2.5.0 version For Linux and MacOS, 
we use UP/DOWN Script for IPv6 block


outspace:
Needs to be tested on all platforms.

PR #238, opened 2023-05-18
Fix macOS build in case QT is installed via brew

sigseg5:


sigseg5:
We can also use multi-threaded mode for `cmake --build . --config release --target all -- -j"$(nproc --all)"` instead of `cmake --build . --config release --target all`. This significantly reduces build time, especially on dev machines.

sigseg5:
Can fix build error with `unset_variable || true`

PR #250, opened 2023-06-08
Fix QT version in deploy script

outspace:
This will fix broken build


outspace:
Looks like git workflow lag. close.

PR #264, opened 2023-07-30
Saving multiple protocols

pokamest:
Saving multiple profiles on iOS


pokamest:
Replaced with https://github.com/amnezia-vpn/amnezia-client/pull/300

PR #308, opened 2023-08-31
Make check_user_in_sudo.sh use sudo -l

Wasapl:
Sudo group is not sufficient in all Linux systems. Correct way is to search for `(ALL) NOPASSWD: ALL` in output of `sudo -l` command.
See https://superuser.com/questions/553932/how-to-check-if-i-have-sudo-access as an example.

### DIFF
--- a/client/core/servercontroller.cpp
+++ b/client/core/servercontroller.cpp
@@ -691,7 +691,7 @@ ErrorCode ServerController::isUserInSudo(const ServerCredentials &credentials, D
     const QString scriptData = amnezia::scriptData(SharedScriptType::check_user_in_sudo);
     ErrorCode error = runScript(credentials, replaceVars(scriptData, genVarsForScript(credentials)), cbReadStdOut, cbReadStdErr);
 
-    if (!stdOut.contains("sudo")) return ErrorCode::ServerUserNotInSudo;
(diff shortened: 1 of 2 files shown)


client/core/servercontroller.cpp
```diff
@@ -691,7 +691,7 @@ ErrorCode ServerController::isUserInSudo(const ServerCredentials &credentials, D
     const QString scriptData = amnezia::scriptData(SharedScriptType::check_user_in_sudo);
     ErrorCode error = runScript(credentials, replaceVars(scriptData, genVarsForScript(credentials)), cbReadStdOut, cbReadStdErr);
 
-    if (!stdOut.contains("sudo")) return ErrorCode::ServerUserNotInSudo;
+    if (!stdOut.contains("(ALL) NOPASSWD: ALL")) return ErrorCode::ServerUserNotInSudo;
```
pokamest, line 694: Using this condition may be not safe
I see output is "(ALL : ALL) NOPASSWD: ALL" on debian

Instead of that we going to send user password to sudo if it asking for it, we will implement it later

pokamest: So, any such fix requires deep testing

pokamest (changes requested):

PR #328, opened 2023-09-16
added close button when drawer shown in the topright corner

ronoaer:
1. show it when open a drawer, and property needCloseButton is true
2. close all of drawers when click top-close-button
3. it will be invisible when there is not drawers

4. todo: it needs be refectory-ed to a single Controller, separated from page-controller

### DIFF
--- a/client/ui/controllers/pageController.h
+++ b/client/ui/controllers/pageController.h
@@ -118,7 +118,7 @@ private:
     std::shared_ptr<Settings> m_settings;
 
     PageLoader::PageEnum m_currentRootPage;
-    int m_drwaerLayer;
+    int m_drawerLayer;
 };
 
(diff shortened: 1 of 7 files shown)


client/ui/controllers/pageController.h
```diff
@@ -104,10 +109,16 @@ public slots:
...
+    void showTopCloseButton(bool visible);
+    void forceCloseDrawer();
+
 private:
     QSharedPointer<ServersModel> m_serversModel;
 
     std::shared_ptr<Settings> m_settings;
+
+    PageLoader::PageEnum m_currentRootPage;
+    int m_drwaerLayer;
```
Nethius, line 121: please correct the variable name to m_drawerLayer

PR #368, opened 2023-10-13
API support

pokamest:


client/vpnconnection.cpp
```diff
@@ -10,7 +10,7 @@
 #include <configurators/shadowsocks_configurator.h>
 #include <configurators/vpn_configurator.h>
 #include <configurators/wireguard_configurator.h>
-#include <core/servercontroller.h>
+#include "core/controllers/serverController.h"
```
tiaga, line 13: Seems this typo leads to failed build and `<` `>` should be used.

client/protocols/openvpnprotocol.cpp
```diff
@@ -214,7 +214,7 @@ ErrorCode OpenVpnProtocol::start()
     m_openVpnProcess->setProgram(PermittedProcess::OpenVPN);
     QStringList arguments({
             "--config", configPath(), "--management", m_managementHost, QString::number(mgmtPort),
-            "--management-client" /*, "--log", vpnLogFileNamePath */
+            "--management-client" , "--log", "/Users/nethius/Documents/git/log"
```
useribs, line 217: Looks suspicious, by mistake?

client/ui/controllers/importController.cpp
```diff
@@ -165,7 +167,11 @@ QJsonObject ImportController::extractAmneziaConfig(QString &data)
         ba = ba_uncompressed;
     }
 
-    return QJsonDocument::fromJson(ba).object();
+    QJsonObject config = QJsonDocument::fromJson(ba).object();
+
+    qDebug() << config;
```
outspace, line 172: please don't write config into the log file, it can contain sensitive data

Noxius87 (changes requested):
Protocols are not switched in the protocol selection menu when switching from the main screen

Nethius:
> Protocols are not switched in the protocol selection menu when switching from the main screen

fixed

Noxius87 (approved):

PR #370, opened 2023-10-14
added new drawer2type for replacing drawertype

ronoaer:


ronoaer:
TODO:
1. adapt DrowDownType
2. adapt PageHome
3. auto hidden
4. block parent mouse event
5. update drawer size

PR #409, opened 2023-10-30
Update servercontroller.cpp, replace 2 calls (shred ; rm)  with one (…

useribs:
…shred -u)


pokamest:
Is this util `shred` available on all Linux distros? The second `rm` was like fallback.
Do we need to check if `shred` is installed then?

useribs:
Nope, it's part of coreutils, like touch. Available even on Solaris, just checked it today )

PR #420, opened 2023-11-05
Improve logic of install_docker.sh

lunardunno:
- check packages update only when it's required
- avoid `dnf/yum update` for RHEL-based systems

### DIFF
--- a/client/server_scripts/install_docker.sh
+++ b/client/server_scripts/install_docker.sh
@@ -1,19 +1,19 @@
-if which apt-get > /dev/null 2>&1; then pm=$(which apt-get); docker_pkg="docker.io"; dist="debian";\
-elif which dnf > /dev/null 2>&1; then pm=$(which dnf); docker_pkg="docker"; dist="fedora";\
-elif which yum > /dev/null 2>&1; then pm=$(which yum); docker_pkg="docker"; dist="centos";\
+if which apt-get > /dev/null 2>&1; then pm=$(which apt-get); silent_inst="-yq install"; check_pkgs="-yq update"; docker_pkg="docker.io"; dist="debian";\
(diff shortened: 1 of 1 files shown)


client/server_scripts/install_docker.sh
```diff
@@ -1,18 +1,19 @@
...
+if ! command -v docker > /dev/null 2>&1; then sudo $pm $silent_inst $docker_pkg;\
   if [ "$dist" = "fedora" ] || [ "$dist" = "centos" ] || [ "$dist" = "debian" ]; then sudo systemctl enable docker && sudo systemctl start docker; fi;\
 fi;\
 if [ "$dist" = "debian" ]; then \
   docker_service=$(systemctl list-units --full --all | grep docker.service | grep -v inactive | grep -v dead | grep -v failed);\
-  if [ -z "$docker_service" ]; then sudo $pm update -yq; sudo $pm install -yq curl $docker_pkg; fi;\
+  if [ -z "$docker_service" ]; then sudo $pm $silent_up; sudo $pm $silent_inst curl $docker_pkg; fi;\
   sleep 3 && sudo systemctl start docker && sleep 3;\
 fi;\
 if ! command -v sudo > /dev/null 2>&1; then echo "Failed to install Docker";exit 1;fi;\
```
useribs, line 19: Why command does not match error message? May be last check should be for docker, not sudo?

lunardunno: Minimal changes were made to this part of the script: only replacing commands with variables.
To minimize the likelihood of new errors appearing.
This part of the script should be analyzed and reworked in the next logical group of changes related to checking the status of docker.service for all values ​​of the $dist variable, in cases where, before using AmneziaVPN, docker was installed on the server, but for some reason, it is inactive or does not start automatically when the server boots while this script is running.

tiaga: Out of scope for the current PR, probably should be fixed further.

tiaga (approved):

pokamest (approved):

PR #453, opened 2023-12-05
added conversion using the system locale

Nethius:


useribs:
Googling revealed there is utf8 method. isn't it a more versatile option? https://doc.qt.io/qt-6/qstring.html#toUtf8

Nethius:
> Googling revealed there is utf8 method. isn't it a more versatile option? https://doc.qt.io/qt-6/qstring.html#toUtf8

I tried toUtf8(), but when using this method, the application keeps crashing

Noxius87 (changes requested):
There's a problem. On MacOS Catalina 10.15 containers are not connected after installation

Noxius87 (changes requested):
AWG protocol does not connect after installation, when trying to install another protocol the application crashes on Windows 11

PR #461, opened 2023-12-14
added escape key handler

Nethius:


Noxius87 (changes requested):
The application crashes when you press Esc (several times) while installing a container or distributing a config

Noxius87 (approved):

pokamest (approved):

PR #462, opened 2023-12-14
fixed server search for sftp, dns and tor containers

Nethius:


Noxius87 (approved):
All installed services are added to the application from the server

PR #463, opened 2023-12-14
ArchLinux_support 

lunardunno:


client/server_scripts/install_docker.sh
```diff
@@ -1,6 +1,7 @@
 if which apt-get > /dev/null 2>&1; then pm=$(which apt-get); silent_inst="-yq install"; check_pkgs="-yq update"; docker_pkg="docker.io"; dist="debian";\
 elif which dnf > /dev/null 2>&1; then pm=$(which dnf); silent_inst="-yq install"; check_pkgs="-yq check-update"; docker_pkg="docker"; dist="fedora";\
 elif which yum > /dev/null 2>&1; then pm=$(which yum); silent_inst="-y -q install"; check_pkgs="-y -q check-update"; docker_pkg="docker"; dist="centos";\
+elif which pacman > /dev/null 2>&1; then pm=$(which pacman); silent_inst="-S --noconfirm --needed --noprogressbar --quiet"; check_pkgs="> /dev/null 2>&1"; docker_pkg="docker"; dist="archlinux";\
```
tiaga, line 4:
```suggestion
elif which pacman > /dev/null 2>&1; then pm=$(which pacman); silent_inst="-S --refresh --noconfirm --needed --quiet"; check_pkgs="> /dev/null 2>&1"; docker_pkg="docker"; dist="archlinux";\
```
1. Seems `--quiet` will be enough and  `--noprogressbar` is not required here.
2. Let's refresh package cache on every install by adding `--refresh`

tiaga:
Arch Linux is not a server OS due to it's rolling-release update cycle thus we shouldn't support it. I'm proposing to create a corresponding issue on GitHub for gathering interest about this feature.

PR #479, opened 2023-12-24
added migration from version 3 of client management to version 4

Nethius:


Noxius87 (changes requested):
There is a problem with porting usernames for Cloak and ShadowSocks protocols

Noxius87 (approved):

PR #480, opened 2023-12-24
added error code output

Nethius:


client/core/errorstrings.cpp
```diff
@@ -2,70 +2,74 @@
...
+QString errorString(ErrorCode code) {
+    QString errorMessage;
+
     switch (code) {
 
     // General error codes
-    case(NoError): return QObject::tr("No error");
-    case(UnknownError): return QObject::tr("Unknown Error");
-    case(NotImplementedError): return QObject::tr("Function not implemented");
+    case(NoError): errorMessage = QObject::tr("No error");
```
albexk: It is necessary to add 'break' to each 'case', otherwise 'switch' will be executed to the end and all error messages will have the text "Internal error"

PR #494, opened 2024-01-12
Xray with Reality protocol

outspace:


client/core/controllers/serverController.cpp
```diff
@@ -574,6 +568,10 @@ ServerController::Vars ServerController::genVarsForScript(const ServerCredential
     vars.append({ { "$FAKE_WEB_SITE_ADDRESS",
                     cloakConfig.value(config_key::site).toString(protocols::cloak::defaultRedirSite) } });
 
+    // Xray vars
+    vars.append({ { "$XRAY_SITE_NAME",
+                  cloakConfig.value(config_key::site).toString(protocols::xray::defaultSite) } });
```
pokamest: Looks like error here - cloakConfig.value for XRay

client/utilities.h
```diff
@@ -26,6 +26,7 @@ class Utils : public QObject
     static QString getStringBetween(const QString &s, const QString &a, const QString &b);
     static bool checkIPv4Format(const QString &ip);
     static bool checkIpSubnetFormat(const QString &ip);
+    static QString getgatewayandiface();
```
pokamest: It's better to give correct name to this function - getGatewayAndIface

service/server/router_linux.cpp
```diff
@@ -102,10 +107,10 @@ bool RouterLinux::routeDelete(const QString &ipWithSubnet, const QString &gw, co
         return false;
     }
 
-    if (ip == "0.0.0.0") {
-        qDebug().noquote() << "Warning, trying to remove default route, skipping: " << ip << gw;
-        return true;
-    }
+ //   if (ip == "0.0.0.0") {
```
pokamest: What are cases to remove default router?

client/containers/containers_defs.cpp
```diff
@@ -85,6 +87,7 @@ QMap<DockerContainer, QString> ContainerProps::containerHumanNames()
              { DockerContainer::Cloak, "OpenVPN over Cloak" },
              { DockerContainer::WireGuard, "WireGuard" },
              { DockerContainer::Awg, "AmneziaWG" },
+             { DockerContainer::Xray, "Xray" },
```
pokamest: We decided to set display name "XRay" everywhere, please fix to XRay

client/utilities.cpp
```diff
@@ -172,6 +203,561 @@ bool Utils::checkIpSubnetFormat(const QString &ip)
...
+        if (pAdapterAddresses) {
+            free(pAdapterAddresses);
+            pAdapterAddresses = nullptr;
+        }
+    }
+
+    return dwRetVal;
+}
+
+static const char* inet_ntop_v4(const void* src, char* dst, socklen_t size) {
```
pokamest: You can replace this implementation by using QHostAddress 
No need this boilerplate to parse IP address to string

client/utilities.cpp
```diff
@@ -172,6 +203,561 @@ bool Utils::checkIpSubnetFormat(const QString &ip)
...
+        *dst++ = digits[n];
+        if (i != 3)
+            *dst++ = '.';
+    }
+    *dst++ = '\0';
+    return orig_dst;
+}
+
+// Helper function for inet_ntop for IPv6 addresses.
+static const char* inet_ntop_v6(const void* src, char* dst, socklen_t size) {
```
pokamest: You can replace this implementation by using QHostAddress 
No need this boilerplate to parse IP address to string

client/utilities.cpp
```diff
@@ -172,6 +203,561 @@ bool Utils::checkIpSubnetFormat(const QString &ip)
...
+        return socket_inet_pton6(src, dst);
+    default:
+        // xerror("EAFNOSUPPORT");
+        return 0;
+    }
+}
+#endif
+
+
+QString Utils::getgatewayandiface()
```
pokamest: You can replace this code with something like this:
  ```
 QList<QNetworkInterface> interfaces = QNetworkInterface::allInterfaces();
    for (const QNetworkInterface &interface : interfaces) {
        if (interface.flags().testFlag(QNetworkInterface::IsUp) &&
            interface.flags().testFlag(QNetworkInterface::CanBroadcast) &&
            !interface.flags().testFlag(QNetworkInterface::IsLoopBack)) {
            for (const QNetworkAddressEntry &entry : interface.addressEntries()) {
                if (entry.ip().protocol() == QAbstractSocket::IPv4Protocol) {
```

File: client/protocols/xrayprotocol.cpp [hunk 8]
```diff
@@ -0,0 +1,243 @@
...
+#include <QJsonObject>
+#include <QNetworkInterface>
+
+
+XrayProtocol::XrayProtocol(const QJsonObject &configuration, QObject *parent):
+    VpnProtocol(configuration, parent)
+{
+    readXrayConfiguration(configuration);
+    m_routeGateway = Utils::getgatewayandiface();
+    m_vpnGateway = "10.33.0.2";
```
pokamest: Please don't hardcode this string in multiple places, it's better to make this address configurable in the proto settings

client/protocols/xrayprotocol.cpp
```diff
@@ -0,0 +1,243 @@
...
+    m_t2sProcess->setProgram(PermittedProcess::Tun2Socks);
+#ifdef Q_OS_WIN
+    QStringList arguments({"-device", "tun://tun2", "-proxy", XrayConStr, "-tun-post-up",
+                           QString("cmd /c netsh interface ip set address name=\"tun2\" static 10.33.0.2 255.255.255.255")});
+#endif
+#ifdef Q_OS_LINUX
+    QStringList arguments({"-device", "tun://tun2", "-proxy", XrayConStr});
+#endif
+#ifdef Q_OS_MAC
+    QStringList arguments({"-device", "utun22", "-proxy", XrayConStr});
```
pokamest, line 122: Why utun22 for macos and tun2 for other OS?

pokamest: Is it possible to iterate tun devices and use first available?

client/protocols/xrayprotocol.cpp
```diff
@@ -0,0 +1,243 @@
...
+    m_routeGateway = Utils::getgatewayandiface();
+    m_vpnGateway = "10.33.0.2";
+    m_vpnLocalAddress = "10.33.0.2";
+}
+
+XrayProtocol::~XrayProtocol()
+{
+    XrayProtocol::stop();
+    QThread::msleep(200);
+#ifndef Q_OS_IOS
```
pokamest: This file shouldn't be compiled for iOS and Android at all, please remove `#ifndef Q_OS_IOS` and avoid compiling this file for mobiles

client/protocols/xrayprotocol.cpp
```diff
@@ -0,0 +1,243 @@
...
+#else
+    return Utils::executable(QString("xray"), true);
+#endif
+}
+
+void XrayProtocol::readXrayConfiguration(const QJsonObject &configuration)
+{
+    m_configData = configuration;
+    QJsonObject xrayConfiguration = configuration.value(ProtocolProps::key_proto_config_data(Proto::Xray)).toObject();
+    int localPort = 10808;
```
pokamest: It's already defined by `defaultLocalProxyPort`, please use it

client/server_scripts/xray/template.json
```diff
@@ -0,0 +1,124 @@
...
+    "log": {
+        "loglevel": "error"
+    },
+    "routing": {
+        "domainStrategy": "IPIfNonMatch",
+        "rules": [
+            {
+                "type": "field",
+                "domain": [
+                    "geosite:category-ads-all"
```
pokamest: Do we need by default block something?

service/server/router_linux.cpp
```diff
@@ -170,3 +175,122 @@ void RouterLinux::flushDns()
     else
         qDebug().noquote() << "OUTPUT systemctl restart nscd/systemd-resolved: " + output;
 }
+
+bool RouterLinux::createTun(const QString &dev, const QString &subnet) {
+    qDebug().noquote() << "createTun start";
+
+    char cmd [1000] = {0x0};
+    sprintf(cmd, "ip tuntap add mode tun dev %s", dev.toStdString().c_str());
+    int sys = system(cmd);
```
pokamest: system() creates new shell process for executing commands.
It's not safe to use, bc it may stuck service.
It's better to rewrite using QProcess, like
  ```
 QProcess process;
    QStringList commands;

    commands << "ip" << "tuntap" << "add" << "mode" << "tun" << "dev" << dev;
    process.start("sudo", commands);
```
and wait some time for process finish

File: service/server/router_linux.cpp [hunk 14]
```diff
@@ -170,3 +175,122 @@ void RouterLinux::flushDns()
...
+    if(sys < 0)
+    {
+        qDebug().noquote() << "Could not activate tun device!\n";
+        return false;
+    }
+
+    return true;
+}
+
+bool RouterLinux::deleteTun(const QString &dev)
```
pokamest, line 230: may be just use `ip link delete <device>` ?

pokamest (changes requested):

client/utilities.h
outspace: fixed

client/core/controllers/serverController.cpp
outspace: fixed

service/server/router_linux.cpp
outspace: This condition is not about default route.  
We need to remove tun2socks routes after disconnecting. This route can be with "0.0.0.0" ip. 

Changed this condition to case with default system route (with mask /0).

client/containers/containers_defs.cpp
outspace: fixed

client/utilities.cpp
outspace: This code is needed only for systems below Windows 7 version. For Windows 7 and greater we can use system function inet_ntop. 
We decided to clean up this code because we don't plan to support systems below Win 7.

client/protocols/xrayprotocol.cpp
outspace: fixed.

client/utilities.cpp
outspace: I'm not sure that we will get a real default GW with that approach.

This approach gives us all system GW but we need to find GW that is currently used.

client/protocols/xrayprotocol.cpp
outspace: fixed

client/protocols/xrayprotocol.cpp
outspace: fixed

service/server/router_linux.cpp
outspace: We use system functions for tun device deletion. That is safer than shell command execution.

client/protocols/xrayprotocol.cpp
outspace: We use tun2 because it is first part of tun2socks app name. 

MacOS we use utun22 on because MacOS the system already has predefined utuns and we need to choose some free utun device.

pokamest (approved):

PR #508, opened 2024-01-18
Update Dockerfile, always get latest released xray version

useribs:
No need to specify version, always get latest release

### DIFF
--- a/client/server_scripts/xray/Dockerfile
+++ b/client/server_scripts/xray/Dockerfile
@@ -1,8 +1,6 @@
 FROM alpine:3.15
 LABEL maintainer="AmneziaVPN"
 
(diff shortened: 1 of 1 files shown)


outspace:
Hi @useribs!

Thanks for your attention to our project. 
Usually, we use fixed versions of binaries in our Dockerfiles. You can see it in Dockerfiles of other protocols. 
We do it for server compatibility with the client binary file version and server config compatibility with the current server binary file version.

useribs:
@outspace Yeah, i thought about client compatibility issues, on the other hand it should be handled/tested by upstream project itself. But i get it we are not in the ideal world ). IIRC there is geoip db bundled there, would it make sense to at least keep it the latest version?

PR #554, opened 2024-02-01
Block traffic between users via wg0 interface

Daloshka:
It'd be much better if someone add a button to turn it on or off


Daloshka:
I deleted other strings. It works by the way and block traffic between users. They can't see each other. I'd like to add button but I don't know how to do it. This is my first commit

pokamest:
No need to remove $WIREGUARD_SUBNET_IP/$WIREGUARD_SUBNET_CIDR syntax. It's planned to implement changing subnet like for OpenVPN

PR #556, opened 2024-02-04
add navigation using enter + buttons will be clicked if enter (if but…

agalehaga:
Added input navigation to TextFieldWithHeaderType elements - setting focus to the next TextFieldWithHeaderType or button. Also added for buttons is changing the border color if the button is in focus, and processing the Enter and Return button presses (= mouse clicks).


client/ui/qml/Controls2/TextFieldWithHeaderType.qml
```diff
@@ -7,6 +7,8 @@ import "TextTypes"
 Item {
     id: root
 
+    signal sig_next()
```
Nethius: Wouldn't it be better to use https://doc.qt.io/qt-6/qml-qtquick-keynavigation.html to navigate with hotkeys?

client/ui/qml/Controls2/TextFieldWithHeaderType.qml
```diff
@@ -7,6 +7,8 @@ import "TextTypes"
 Item {
     id: root
 
+    //signal sig_next()
```
Nethius: please, cleanup unused code

client/ui/qml/Pages2/PageProtocolOpenVpnSettings.qml
```diff
@@ -403,4 +416,9 @@ PageType {
             id: questionDrawer
         }
     }
+
+    function init()
```
Nethius: I think it's better to use Component.onCompleted:

pokamest (approved):

PR #579, opened 2024-02-12
Migrate to a new AmneziaWG Docker Image

AlexanderGalkov:
Update server scripts


pokamest:
Probably need to provide backward compatibility with older container version, look ClientManagementModel::getWireGuardClients

PR #593, opened 2024-02-14
Setup MTU for iOS

isamnezia:


client/vpnconnection.cpp
```diff
@@ -305,6 +305,11 @@ QJsonObject VpnConnection::createVpnConfiguration(int serverIndex, const ServerC
     // TODO: try to get hostName, port, description for 3rd party configs
     // vpnConfiguration[config_key::port] = ...;
 
+#if defined Q_OS_IOS
+    QJsonObject awg = server[config_key::containers].toArray().first().toObject()[config_key::awg].toObject();
```
outspace: What about WG config? Does it apply to WG config?

isamnezia: There was no mtu field in wg json. And now there is no mtu field in openvpn json.

outspace: For WG config it has same logic to AWG config. Please try to change mtu for WG protocol and check that field.

isamnezia: I know about same logic. I mean wg/awg. I will write to you in tg.

PR #596, opened 2024-02-15
UI fixes

andr13:


client/ui/qml/Pages2/PageStart.qml
```diff
@@ -66,6 +67,7 @@ PageType {
             } else {
                 tabBarStackView.push(pagePath, { "objectName" : pagePath }, StackView.Immediate)
             }
+            tabBar.isServerInfoShow = page === PageEnum.PageSettingsServerInfo
```
Nethius, line 70: From this page we can go on to the various protocols/services configuration pages and protocols/services installation pages. When going to these pages, the "PageHome" icon is highlighted again

Nethius: From this page we can go on to the various protocols/services configuration pages and protocols/services installation pages. When going to these pages, the "PageHome" icon is highlighted again

Nethius (changes requested):

client/ui/qml/Pages2/PageStart.qml
andr13: Ah, right. I will fix it today, thanks

Nethius (approved):

Issue #428, opened 2023-11-16
Install sftp error (and openssl key recognition)


barsikus007:
**Describe the bug**
I tried install Amnezia and got 2 errors
First one was ssh key error, so I enabled password auth and used password
Second one in strange `Sftp error: File does not exist` when I tried to install any of suggested VPNs

**To Reproduce**
- Follow usual install flow
- Any protocol raises same Sftp error

**Expected behavior**
Successful installation

**Log files**
2023-11-16 14:01:33 debug buildContainerWorker start
2023-11-16 14:01:33 critical SFTP server: No such file

**Screenshots**
![image](https://github.com/amnezia-vpn/amnezia-client/assets/37113583/590d7489-7a04-43a6-82a1-a1e03111bd35)

**Desktop (please complete the following information):**
- Windows and andriod version 4.0.8

**Server (please complete the following information):**
- Ubuntu 20.04

**Additional context**
Also, my rsa key (generated by ssh-keygen) won't accept by application
why?
![image](https://github.com/amnezia-vpn/amnezia-client/assets/37113583/ad6ace1e-6a33-4605-a083-b77d4d14cf52)

GeliusIHe:
Same problem with sftp error. However, I get this error only on the Ubuntu server, which is under wsl2. If the main server is Ubuntu, everything is fine.

ANGIPO:
Up, same error. Ubuntu 22.04

KotaruProject:
Подтверждаю, та же ошибка  Ubuntu 22.04.2 LTS

wolfhound234:
Same error. Ubuntu 22.04

barsikus007:
okay, I realized issue

it works ONLY at root
this is unsecure as f

ANGIPO:
solution:
1) open ssh cfg:
sudo nano /etc/ssh/sshd_config

2) uncomment "PermitRootLogin yes" (delete #)
3) set password for root:
passwd root
4) authorization in the amnezia vpn program via root

nkz-soft:
> solution:
> 
>     1. open ssh cfg:
>        sudo nano /etc/ssh/sshd_config
> 
>     2. uncomment "PermitRootLogin yes" (delete #)
> 
>     3. set password for root:
>        passwd root
> 
>     4. authorization in the amnezia vpn program via root

In my opinion, this is a serious security issue. It's best to use a separate user and key instead of logging in as root on your server.

RomikB:
[...]

PR #602, opened 2024-02-16
Replace sftp with scp 

dimov96:
Replacing the libssh::Client::sftpFileCopy  wtih libssh::Client::scpFileCopy. (#428)

### DIFF
--- a/client/configurators/openvpn_configurator.cpp
+++ b/client/configurators/openvpn_configurator.cpp
@@ -76,7 +76,7 @@ OpenVpnConfigurator::ConnectionData OpenVpnConfigurator::prepareOpenVpnConfig(co
 
     if (connData.caCert.isEmpty() || connData.clientCert.isEmpty() || connData.taKey.isEmpty()) {
         if (errorCode)
-            *errorCode = ErrorCode::OpenVpnUnknownError;
+            *errorCode = ErrorCode::SshScpFailureError;
     }
 
     return connData;
--- a/client/configurators/openvpn_configurator.cpp
+++ b/client/configurators/openvpn_configurator.cpp
@@ -76,7 +76,7 @@ OpenVpnConfigurator::ConnectionData OpenVpnConfigurator::prepareOpenVpnConfig(co
 
     if (connData.caCert.isEmpty() || connData.clientCert.isEmpty() || connData.taKey.isEmpty()) {
         if (errorCode)
-            *errorCode = ErrorCode::SshSftpFailureError;
+            *errorCode = ErrorCode::OpenVpnUnknownError;
     }
 
     return connData;
--- a/client/core/defs.h
+++ b/client/core/defs.h
@@ -48,10 +48,6 @@ namespace amnezia
 
         // Ssh scp errors
         SshScpFailureError = 400,
-        SshScpOpenError = 401,
-        SshScpReadError = 402,
-        SshScpPermissionsError = 403,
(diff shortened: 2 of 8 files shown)


Nethius (approved):

client/configurators/openvpn_configurator.cpp
```diff
@@ -76,7 +76,7 @@ OpenVpnConfigurator::ConnectionData OpenVpnConfigurator::prepareOpenVpnConfig(co
 
     if (connData.caCert.isEmpty() || connData.clientCert.isEmpty() || connData.taKey.isEmpty()) {
         if (errorCode)
-            *errorCode = ErrorCode::SshSftpFailureError;
+            *errorCode = ErrorCode::OpenVpnUnknownError;
```
pokamest: OpenVpnUnknownError is not suitable for this case, this error is not related to OpenVPN by itself. It's better to define new error codes if necessary

client/core/sshclient.cpp
```diff
@@ -222,99 +222,89 @@ namespace libssh {
...
-    ErrorCode Client::sftpFileCopy(const SftpOverwriteMode overwriteMode, const QString& localPath, const QString& remotePath, const QString &fileDesc)
+    ErrorCode Client::scpFileCopy(const ScpOverwriteMode overwriteMode, const QString& localPath, const QString& remotePath, const QString &fileDesc)
     {
-        m_sftpSession = sftp_new(m_session);
+        m_scpSession = ssh_scp_new(m_session, SSH_SCP_WRITE, remotePath.toStdString().c_str());
 
-        if (m_sftpSession == nullptr) {
-            return closeSftpSession();
+        if (m_scpSession == nullptr) {
+            return closeScpSession();
```
pokamest: Since we started refactoring, please improve error handling part, add new error codes if necessary.
`closeScpSession()` error code will be returned instead of actual error which we can retrieve using `ssh_get_error`

client/core/sshclient.cpp
```diff
@@ -222,99 +222,89 @@ namespace libssh {
...
 
             fin.close();
 
-            int result = sftp_close(file);
-            if (result != SSH_OK) {
-                return closeSftpSession();
-            }
-
-            return closeSftpSession();
+            return closeScpSession();
```
pokamest: Looks like you need to review and simplify whole scpFileCopy logic.
For example, closeScpSession() missing in some function exit conditions, but QFile fin not necessary to close at all.

pokamest (changes requested):

Nethius (approved):

pokamest (approved):

client/core/defs.h
```diff
@@ -83,7 +79,13 @@ namespace amnezia
...
         // Api errors
         ApiConfigDownloadError = 1100,
-        ApiConfigAlreadyAdded = 1101
+        ApiConfigAlreadyAdded = 1101,
+
+        // QFile errors
+        OpenError = 1200,
+        ReadError = 1201,
+        PermissionsError = 1202,
+        UnspecifiedError = 1203,
```
Yunir, line 88: The documentation does not contain them in https://docs.amnezia.org/troubleshooting/error-codes/
I face 1200 and don't know now how can I resolve it

Yunir: it helped to restart the amnezia-client and try to setup the vpn on vps one more time.

PR #635, opened 2024-02-22
added export awg native format

agalehaga:
added export awg native format + fixed little bugs

### DIFF
--- a/client/ui/controllers/systemController.cpp
+++ b/client/ui/controllers/systemController.cpp
@@ -92,7 +92,7 @@ QString SystemController::getFileName(const QString &acceptLabel, const QString 
 
     mainFileDialog->setProperty("acceptLabel", QVariant::fromValue(acceptLabel));
     mainFileDialog->setProperty("nameFilters", QVariant::fromValue(QStringList(nameFilter)));
-    if (QFileInfo::exists(selectedFile)) {
+    if (!selectedFile.isEmpty()) {
         mainFileDialog->setProperty("selectedFile", QVariant::fromValue(selectedFile));
     }
     mainFileDialog->setProperty("isSaveMode", QVariant::fromValue(isSaveMode));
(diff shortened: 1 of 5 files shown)


client/ui/qml/Pages2/PageShare.qml
```diff
@@ -63,6 +67,13 @@ PageType {
                 shareConnectionDrawer.configFileName = "amnezia_for_wireguard"
                 break
             }
+            case PageShare.ConfigType.Awg: {
+                ExportController.generateAwgConfig(clientNameTextField.textFieldText)
+                shareConnectionDrawer.configCaption = qsTr("Save Awg config")
```
Nethius: please change to AmneziaWG

client/ui/qml/Pages2/PageShare.qml
```diff
@@ -110,6 +121,11 @@ PageType {
         property string name: qsTr("WireGuard native format")
         property var type: PageShare.ConfigType.WireGuard
     }
+    QtObject {
+        id: awgConnectionFormat
+        property string name: qsTr("Awg native format")
```
Nethius: please change to AmneziaWG

client/ui/controllers/systemController.cpp
```diff
@@ -92,7 +92,7 @@ QString SystemController::getFileName(const QString &acceptLabel, const QString
 
     mainFileDialog->setProperty("acceptLabel", QVariant::fromValue(acceptLabel));
     mainFileDialog->setProperty("nameFilters", QVariant::fromValue(QStringList(nameFilter)));
-    if (!selectedFile.isEmpty()) {
+    if (QFileInfo::exists(selectedFile)) {
```
Nethius: please return check for an empty string

Nethius (changes requested):

client/ui/controllers/systemController.cpp
agalehaga: On my computer the file selected is === C:\Users\dvajd\OneDrive\Documents\amnezia.conf (or something like that). I do not have such a file (maybe there is a problem with the Cyrillic alphabet on the way) - it is why I have qml error - qrc:/ui/qml/main2.qml:239:5: QML FileDialog: Cannot set C:/Users/dvajd/OneDrive/?????????/amnezia_for_awg as a selected file because it doesn't exist (error 2). What correct way fix it?

Nethius: This line is not to check if the file exists, because if the file does not exist, it must be created. The error you are talking about is fixed here https://github.com/amnezia-vpn/amnezia-client/pull/636

Nethius (approved):

PR #656, opened 2024-02-29
resetting QR page content when generating a QR code after the first time

ShehabAhmed-Gui:
fixed the first-generated QR code is visible while generating another QR code bug


Nethius (changes requested):
I think adding emit exportConfigChanged() to the end of clearPreviousConfig() would solve the problem

Nethius (approved):

pokamest (approved):

PR #682, opened 2024-03-11
Fix UI for Burmese language

andr13:


client/ui/qml/Controls2/TextTypes/ParagraphTextType.qml
```diff
@@ -1,7 +1,7 @@
 import QtQuick
 
 Text {
-    lineHeight: 24
+    lineHeight: LanguageModel.getCurrentLanguageIndex() === 5 ? 30 : 24 // Burmese
```
Nethius: it is better to make a property in languageModel, which will return a value depending on the selected language

Nethius (changes requested):

client/ui/qml/Controls2/TextTypes/ParagraphTextType.qml
andr13: you are right, that would be better, I'll fix it

andr13:
#689

Nethius (approved):

pokamest (approved):

PR #707, opened 2024-03-21
Disabled button redesign

andr13:


andr13:
Maybe we should experiment with colors some more

PR #716, opened 2024-03-28
Add checking background service before connect

fameowner99:
Just check if background proccess is alive.
Also this method will work in linux so we can extend to check also on linux.
Didnt find working objective-c way to check background services permissions


client/vpnconnection.cpp
```diff
@@ -316,6 +323,17 @@ QJsonObject VpnConnection::createVpnConfiguration(int serverIndex, const ServerC
 void VpnConnection::connectToVpn(int serverIndex, const ServerCredentials &credentials, DockerContainer container,
                                  const QJsonObject &containerConfig)
 {
+    m_connectionError.reset();
+
+#ifdef Q_OS_APPLE
+    if (!MacOSUtils::isBackgroundServicesEnabled())
```
Nethius: please move this check to ConnectionController

Nethius (changes requested):

lunardunno (approved):

Nethius (approved):

pokamest (approved):

PR #749, opened 2024-04-10
Prevent editing active container

fameowner99:


client/ui/models/servers_model.cpp
```diff
@@ -386,7 +386,7 @@ void ServersModel::updateContainerConfig(const int containerIndex, const QJsonOb
     server.insert(config_key::containers, containers);
 
     auto defaultContainer = server.value(config_key::defaultContainer).toString();
-    if ((ContainerProps::containerFromString(defaultContainer) == DockerContainer::None
+    if (!m_isConnectionEnabled && (ContainerProps::containerFromString(defaultContainer) == DockerContainer::None
          || ContainerProps::containerService(container) != ServiceType::Other)) {
         server.insert(config_key::defaultContainer, ContainerProps::containerToString(container));
     }
```
fameowner99: This code should be executed only if no active connection.
With active connection defaultContainer will be change so it will lead to a bug.
I dont't know how to check if connection is active in ServersModel,
for now I see two options:
1) to use signal slot like in current pull request
2) create member pointer ConnectionContoller in ServersModel class and access ConnectionContoller::isConnected() const function  directly

lunardunno (approved):

Noxius87 (approved):

lunardunno (approved):

PR #765, opened 2024-04-19
Fix Android routes processing

albexk:


albexk:
The changes of this PR were merged as part of #797

PR #770, opened 2024-04-21
Censorship levels translation update

andr13:


andr13:
Translation update for #768

PR #784, opened 2024-04-28
Fix opening url after save config

fameowner99:


fameowner99:
Before fix I had such error 
image

Noxius87 (approved):

pokamest (approved):

PR #794, opened 2024-05-05
Request internet permission before connect for iOS

pokamest:


lunardunno (approved):
Windows, Linux, Andoid, All ок.

PR #795, opened 2024-05-05
Fdroid metadata

ShehabAhmed-Gui:
add F-droid metadata


metadata/en-US/changelogs/44000.txt
albexk, line 1: It is necessary to delete this file. I don't think we will release version 4.4.0 on fdroid.

metadata/en-US/changelogs/45300.txt
albexk, line 1: It is necessary to rename this file. The current versionCode is in `/CMakeLists.txt`: `set(APP_ANDROID_VERSION_CODE 52)`. Please also note this in the build metadata file. And versionName matches the version of the project (4.5.3.0).

metadata/en-US/changelogs/45300.txt
```diff
@@ -0,0 +1,5 @@
+What's Changed:
+* WireGuard obfuscation feature added
```
albexk: The changelog needs to be replaced. Android changelog is different from the changelog in the release on github. The current android changelog is available from google play.

metadata/en-US/full_description.txt
```diff
@@ -0,0 +1,22 @@
...
+
+Free open-source software to create a personal VPN on your server. Helps to access blocked content without revealing privacy even to VPN providers. Specify IP address, credentials for your server and Amnezia will configure it to connect via VPN automatically.
+
+We are not a VPN service, you don't have to connect to our servers and pay us anything. You free to use your own self-hosted VPS or you could purchase any VPS from any provider. To connect, use your own or purchase any VPS from any provider.
+
+Features
+
+• Connect using a config file, text code, by scanning a QR code, or setting up your own server
+• Select and set the protocols manually or use the "Wizard" function to configure them automatically based on the level of censorship in your country.
+• Supports the following VPN protocols on the server: WireGuard, AmneziaWG, OpenVPN, OpenVPN over Cloak, ShadowSocks and IKEv2.
```
albexk: ShadowSocks and IKEv2 are not supported on Android, so it's better to remove them from this list. @pokamest, or is it better to completely copy the description from google play?

metadata/en-US/title.txt
```diff
@@ -0,0 +1 @@
+AmneziaVPN - Multi protocol VPN client
```
albexk: Perhaps here it is better to take the name also from google play: just `Amnezia VPN`.

albexk (changes requested):

metadata/en-US/full_description.txt
pokamest: Agree

albexk (approved):

Issue #832, opened 2024-05-30
Amnezia windows client seems unstable when using AmneziaWG protocol


just-dont:
**Describe the bug**
If I'm using AmneziaWG, then inevitably after at least one reconnection event something breaks bad enough to mess all the programs and services requiring localhost loopback, necessitating PC reboot.
The service log is filled with thousands of the following lines, repeating endlessly:

2024-05-28 18:59:46 debug Amnezia "WindowsSplitTunnel" "No Split-Tunnel Conflict detected, continue."
2024-05-28 18:59:46 debug Amnezia "WindowsSplitTunnel" "Try to open Split Tunnel Driver"
2024-05-28 18:59:46 debug Amnezia "WindowsUtils" "Failed to open Driver:  - Access is denied."
2024-05-28 18:59:46 debug Amnezia "WindowsServiceManager" "OpenSCManager access given -  0"
2024-05-28 18:59:46 debug Amnezia "WindowsServiceManager" "Opening Service -  AmneziaVPNSplitTunnel"
2024-05-28 18:59:46 debug Amnezia "WindowsServiceManager" "Service manager execute access granted"
2024-05-28 18:59:46 debug Amnezia "WindowsServiceManager" "Service start not possible, as its running"

It doesn't happen if I use any other service. It doesn't happen if I use AmneziaWG but connect only once after rebooting, never turning VPN off and never reconnecting again.
I also tried to run it with Split Tunneling on (both by host and by app), on only by host, and off. This doesn't seem to change anything about this issue.

There's one interesting line in the logs, repeating occasionally, prior to the stuff quoted above:
`2024-05-29 09:48:14 debug Amnezia "WindowsSplitTunnel" "Stopping Split tunnel not successfull"`

What's different with the other similar issues I found is that even when the issue is already happening and e.g. my VMWare software isn't working properly (since it depends on the localhost availability), the VPN still technically works.

**To Reproduce**
1. Connect amnezia client using AmneziaWG protocol.
2. Disconnect and reconnect at least once.
3. After some time period, observe the problem via logs and failed localhost connections.

**Expected behavior**
Amnezia shouldn't affect my PC at least when turned off.

**Log files**
[AmneziaVPN.log](https://github.com/amnezia-vpn/amnezia-client/files/15495629/AmneziaVPN.log)
[AmneziaVPN-service.log](https://github.com/amnezia-vpn/amnezia-client/files/15495630/AmneziaVPN-service.log)

[...]

PR #871, opened 2024-06-28
Do not allow to add loopback/multicast/broadcast ips to split tunnel

fameowner99:


fameowner99:
Related to https://github.com/amnezia-vpn/amnezia-client/issues/832

PR #872, opened 2024-06-28
Fix TorWebsite setup in UI

outspace:
Fixes torwebsite container setup from UI. 
Without this fix "Install" button just do nothing for Tor container.


lunardunno (approved):
Now the TorWebSite install button works fine.

PR #857, opened 2024-06-18
update Russian translation

StrikerRUS:

PR #891, opened 2024-07-12
Rename Shadowsocks

StrikerRUS:
Refer to https://github.com/amnezia-vpn/amnezia-client/pull/857#discussion_r1645216307.


client/containers/containers_defs.cpp
```diff
@@ -90,13 +90,13 @@ QMap<DockerContainer, QString> ContainerProps::containerHumanNames()
 {
     return { { DockerContainer::None, "Not installed" },
              { DockerContainer::OpenVpn, "OpenVPN" },
-             { DockerContainer::ShadowSocks, "OpenVPN over SS" },
+             { DockerContainer::ShadowSocks, "OpenVPN over Shadowsocks" },
```
lunardunno: If rename OpenVPN over SS to OpenVPN over Shadowsocks, the interface in this place may look bad again.
image

StrikerRUS: Ok, makes sense! I'll revert this particular change.

StrikerRUS:
```suggestion
             { DockerContainer::ShadowSocks, "OpenVPN over SS" },
```

PR #898, opened 2024-07-16
Im gs for readme

KsZnak:
Changes have been made to the readme, download buttons and a banner have been added. Fixed cheats for donations.

PR #905, opened 2024-07-21
Fix README image paths

activivan:
Fixing image paths in README.md

#898


pokamest:
Was fixed in another PR

PR #1045, opened 2024-09-04
GoodByeDPI support

outspace:


StrikerRUS:
Hey, great idea to add a DPI bypass tool into Amnezia!
But I believe that nowadays [`Zapret`](https://github.com/bol-van/zapret) is a more feature rich, cross platform and actively maintained tool. Also, it will be good to integrate something like [`GoodCheck`](https://ntc.party/t/goodcheck-%D0%B1%D0%BB%D0%BE%D0%BA%D1%87%D0%B5%D0%BA-%D1%81%D0%BA%D1%80%D0%B8%D0%BF%D1%82-%D0%B4%D0%BB%D1%8F-goodbyedpi-zapret-byedpi/10880) with the aim to provide users a good enough initial config for their ISP, and keep Amnezia user-friendly with low entry barrier.

PR #1141, opened 2024-10-09
[draft] added windows onlyForwardApps selection

mayoshia:
researching the problem


Nethius:
Just because you have unlocked the selector, this mode will not work on windows. This functionality is not implemented in windows

mayoshia:
> t

yeah i know, im just trying to compile it via workflow and research a little bit further

PR #1237, opened 2024-11-07
feature/mozilla upstream

Nethius:


client/platforms/windows/daemon/windowsfirewall.cpp
```diff
@@ -236,17 +262,6 @@ bool WindowsFirewall::enablePeerTraffic(const InterfaceConfig& config) {
     }
   }
 
-  if (!config.m_excludedAddresses.empty()) {
```
outspace, line 239: We need this part for splittunnel excluded sites

client/platforms/windows/daemon/windowsfirewall.cpp
```diff
@@ -484,57 +549,6 @@ bool WindowsFirewall::allowTrafficTo(const QHostAddress& targetIP, uint port,
   return true;
 }
 
-bool WindowsFirewall::allowTrafficToRange(const IPAddress& addr, uint8_t weight,
```
outspace, line 487: Split tunnel supports routes with network mask. That's why we rewrite default "allowTraffic".

client/platforms/windows/daemon/windowssplittunnel.cpp
```diff
@@ -273,58 +443,60 @@ std::vector<uint8_t> WindowsSplitTunnel::generateAppConfiguration(
...
-      }
-      break;
-    }
+  if (ipv6.empty()) {
+    std::memset(out_ipv6, 0x00, sizeof(IN6_ADDR));
+    return true;
+  }
+  if (InetPtonW(AF_INET6, ipv6.c_str(), out_ipv6) != 1) {
+    logger.debug() << "Ipv6 Conversation error" << WSAGetLastError();
+    return false;
```
outspace: Not sure about this part. It breaks app split tunnel when ipv6 is not setup. 

We can just ignore when ipv6 address is not setup.

pokamest (approved):

PR #1262, opened 2024-11-21
bugfix: added scroll on page with services list

Nethius:


client/ui/qml/Pages2/PageSetupWizardApiServicesList.qml
```diff
@@ -16,83 +16,88 @@ PageType {
...
+        anchors.bottomMargin: 16
+        spacing: 16
 
-                headerText: qsTr("VPN by Amnezia")
-                descriptionText: qsTr("Choose a VPN service that suits your needs.")
-            }
+        currentIndex: 1
+        clip: true
+        model: ApiServicesModel
+
```
aiamnezia, line 63:
```suggestion
reuseItems: true
```

Can reuse objects for delegates, if some of them are unseen and others are apearing.

client/ui/qml/Pages2/PageSetupWizardApiServicesList.qml
```diff
@@ -16,83 +16,88 @@ PageType {
...
+            policy: servicesListView.height >= servicesListView.contentHeight ? ScrollBar.AlwaysOff : ScrollBar.AlwaysOn
+        }
 
-            ListView {
-                id: containers
-                width: parent.width
-                height: containers.contentItem.height
-                spacing: 16
+        Keys.onUpPressed: scrollBar.decrease()
+        Keys.onDownPressed: scrollBar.increase()
```
aiamnezia: AFAIK ListView already provides built-in keyboard navigation support, so these handlers might be redundant unless there's a specific requirement for custom scrolling behavior.

Are you sure that this explicit custom implementation is necessary?

I don't know, maybe you don't want to change currentIndex and default behaviour change it

client/ui/qml/Pages2/PageSetupWizardApiServicesList.qml
```diff
@@ -16,83 +16,88 @@ PageType {
...
-                descriptionText: qsTr("Choose a VPN service that suits your needs.")
-            }
+        currentIndex: 1
+        clip: true
+        model: ApiServicesModel
+
+        ScrollBar.vertical: ScrollBar {
+            id: scrollBar
+            policy: servicesListView.height >= servicesListView.contentHeight ? ScrollBar.AlwaysOff : ScrollBar.AlwaysOn
+        }
```
aiamnezia, line 68: While this provides explicit control, the default `ScrollBar.AsNeeded` policy might be more suitable here as it handles visibility automatically with smooth transitions.

Also, sometimes a saw that the default behavior of ScrollBar is not working correctly. To fix this, you can initialize the ScrollBar as an empty object like this. 

```suggestion
        ScrollBar.vertical: ScrollBar {}
```
Qt will then set the necessary default values automatically.

aiamnezia: While this provides explicit control, the default `ScrollBar.AsNeeded` policy might be more suitable here as it handles visibility automatically with smooth transitions.

Also, sometimes a saw that the default behavior of ScrollBar is not working correctly. To fix this, you can initialize the ScrollBar as an empty object like this. 

```
    ScrollBar.vertical: ScrollBar {}
```
Qt will then set the necessary default values automatically.

client/ui/qml/Pages2/PageSetupWizardApiServicesList.qml
Nethius: I don't think it needs to be enabled for this listview

aiamnezia (approved):

PR #1331, opened 2024-12-30
bugfix: Correctly use QProcess::start and QProcess::execute

al42and:
Affected functions (all on Linux/Mac):
 - `RouterLinux::flushDns` was not reloading the DNS manager.
 - `Utils::processIsRunning` was always saying that the process is not running when `fullFlag` was set to `false`.
 - `Utils::killProcessByName` was not killing anything.

### DIFF
--- a/client/utilities.cpp
+++ b/client/utilities.cpp
@@ -194,8 +194,13 @@ bool Utils::processIsRunning(const QString &fileName, const bool fullFlag)
     return false;
 #else
     QProcess process;
+    QStringList arguments;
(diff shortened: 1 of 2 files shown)


outspace:
Thanks, but we have problem with Utils::killProcessByName on Linux too. 
Can you fix it?

al42and:
Good catch! I did not check that the old process was actually killed :smile:

Also, found a similar pattern in `RouterLinux::flushDns`.

outspace (approved):

PR #1333, opened 2025-01-01
Killswitch strict mode

### LINKED ISSUE #1252: Feature request. Improvements to the kill switch function.
I used Windscribe vpn before, but due to recent events I decided to try Amnezia. Overall, everything is great, but I noticed the lack of some features that are quite important in my opinion.

The windows desktop client has a kill switch function, but its capabilities seem insufficient to me. As far as I understand, when switching servers, I need to disconnect from the previous server manually, and at this point all traffic from the computer goes without vpn. Most likely, the same thing can happen during the operating system boot before launching the application. It would be great to have the ability to block all traffic while vpn is inactive. For comparison, how it is [done](https://windscribe.com/features/firewall) in windscribe.

I am also interested in the question about [this](https://windscribe.com/features/tcp-socket-termination) feature. Is there something similar in Amnezia and can data leak due to unclosed connections?

outspace:
Killswitch strict mode can be enabled by option "Conf/strictKillSwitchEnabled" in application config.


service/server/killswitch.cpp
```diff
@@ -150,28 +151,49 @@ bool KillSwitch::disableAllTraffic() {
...
 
+bool KillSwitch::addAllowedRange(const QStringList &ranges) {
+    for (const QString &range : ranges) {
+        if (!range.isEmpty() && !m_allowedRanges.contains(range)) {
+            m_allowedRanges.append(range);
+        }
+    }
+
+#ifdef Q_OS_WIN
+    WindowsFirewall::create(this)->allowTrafficRange(ranges);
```
outspace: we have this part in "allowTrafficTo"

service/server/killswitch.cpp
```diff
@@ -150,28 +151,49 @@ bool KillSwitch::disableAllTraffic() {
     MacOSFirewall::setAnchorEnabled(QStringLiteral("000.allowLoopback"), true);
     MacOSFirewall::setAnchorEnabled(QStringLiteral("250.blockIPv6"), true);
 #endif
+    m_allowedRanges.clear();
     return true;
 }
 
 bool KillSwitch::allowTrafficTo(const QStringList &ranges) {
```
outspace: rename it into "resetAllowedRange"

Nethius (approved):

PR #1335, opened 2025-01-01
fix: rewrite linux router dns flusher

sund3RRR:
Rewrite the current linux's router method that finds out which DNS manager is being used from checking the `.service` path to `systemctl list-units`.
Context: https://github.com/NixOS/nixpkgs/pull/367972#discussion_r1900237580

### DIFF
--- a/service/server/router_linux.cpp
+++ b/service/server/router_linux.cpp
@@ -152,21 +152,29 @@ bool RouterLinux::routeDeleteList(const QString &gw, const QStringList &ips)
     return cnt;
 }
 
+bool RouterLinux::isServiceActive(const QString &serviceName) {
+    QProcess process;
+    process.start("systemctl", { "is-active", "--quiet", serviceName });
+    process.waitForFinished();
+
+    return process.exitCode() == 0;
+}
+
 void RouterLinux::flushDns()
 {
     QProcess p;
     p.setProcessChannelMode(QProcess::MergedChannels);
 
     //check what the dns manager use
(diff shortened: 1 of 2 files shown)


service/server/router_linux.cpp
```diff
@@ -152,21 +154,37 @@ bool RouterLinux::routeDeleteList(const QString &gw, const QStringList &ips)
     return cnt;
 }
 
+bool RouterLinux::isServiceActive(const QString &serviceName) {
+    QProcess process;
+    process.start("systemctl", { "list-units", "--type=service", "--all", "--no-legend" });
```
outspace: 1. We need services with "active" status.
2. It can be rewrited using "systemctl is-active --quiet nscd.service" and exit code.

sund3RRR: 1. Added `--state=running` arg, done
2. I think we should keep this implementation since we can easily add other DNS managers in the future.

outspace: This method is not about DNS managers. 
It checks that service in active state by name. Systemctl has default command "[is-active](https://www.man7.org/linux/man-pages/man1/systemctl.1.html)" for this.

Your implementation will work but it obtain full list of active services. I'm not shure why we need it.

Please explain this.

sund3RRR: Hmm, fine, sounds good

outspace (approved):

outspace:
Thanks!

Nethius (approved):

Issue #1201, opened 2024-10-25
Working only in bash shell only


YouROK:
**Describe the bug**
If a default shell other than bash is installed on the server, the installation fails.

**To Reproduce**
On vps server:
```
apt install fish
chsh
# enter "/usr/bin/fish"
```
On host in AmneziaVPN select Self-hosted VPN and all next. VPN protocol I select awg.
And click Install button

**Expected behavior**
Execute scripts in the bash, not in the default shell.

**Screenshots**
<img width="492" alt="image" src="https://github.com/user-attachments/assets/a45d7647-2a79-444e-be86-1ea8ccbb4262">

**Desktop (please complete the following information):**
 - OS: MacOS 15.0.1
 - Version 4.8.1.9

**Server (please complete the following information):**
 - OS: Debian 12

vanyasem:
The scripts should specify a [shebang](https://en.wikipedia.org/wiki/Shebang_%28Unix%29) `#!/usr/bin/env bash` instead of relying on the default shell

[check_user_in_sudo.sh](https://github.com/amnezia-vpn/amnezia-client/blob/4.8.2.3/client/server_scripts/check_user_in_sudo.sh) fails to get executed on fish shell, resulting in an [Error 205: ServerUserNotInSudo](https://docs.amnezia.org/ru/troubleshooting/error-codes/#error-205-serverusernotinsudo):

```
$ CUR_USER=$(whoami);\
  groups $CUR_USER
fish: Unsupported use of '='. In fish, please use 'set CUR_USER $(whoami)'.
```

lunardunno:
1. This script is executed by the program line by line in sh.
2. Not all script error messages cause the server controller to stop.
3. Your first message shows error 206, not 205, because there is something wrong with your server or with the user you created (if you are not using root).
4. Work is underway to improve the check_user_in_sudo.sh script.

vanyasem:
@lunardunno 

You're right, a shebang won't help, as the script is executed line-by-line. My mistake

However, it uses the default user's shell instead of directly calling `sh`:

1) The script is ran in `ServerController::isUserInSudo`:
https://github.com/amnezia-vpn/amnezia-client/blob/157064763e0897e60fd9461bf0635ced38315a6d/client/core/controllers/serverController.cpp#L771

2) `ServerController::runScript` sends the command to an ssh client:
[...]

PR #1442, opened 2025-03-01
Checking server user permissions to use sudo

### LINKED ISSUE #1133: The installer script may not work if user has sudo privileges given with a different user group other than "sudo"
**Describe the bug**
The installer script checks sudo privileges based on the groups given to the user (the group in question named `sudo`). This may conflict if the sudo group in question is named something else (e.g. `wheel`).

**To Reproduce**
The steps to reproduce this problem are hypothetical and are made based on previous experience.

Steps to reproduce the behavior on Arch Linux:
1. Install `sudo` package on the server.
2. [Using one of the example entries add sudo privileges to the `wheel` group.](https://wiki.archlinux.org/title/Sudo#Example_entries)
3. Create `wheel` group.
4. Create a user and assign it to `wheel`.
5. Try to install any VPN server with Amnezia client using the said server and user.
6. See error.

**Expected behavior**
The installer script should work just fine even if the group that is used to give sudo privileges is named `wheel`.

**Server**
 - OS: any Linux distribution

**Additional context**
https://github.com/amnezia-vpn/amnezia-client/blob/d4fff4af3cf836cc021c68f35779759c9152a96f/client/core/controllers/serverController.cpp#L768-L772

https://github.com/amnezia-vpn/amnezia-client/blob/d4fff4af3cf836cc021c68f35779759c9152a96f/client/server_scripts/check_user_in_sudo.sh#L1-L2

This problem can be fixed using `sudo -v` or `sudo -l` in favor of checking user groups.

lunardunno:
Adding support for the wheel group to work with the sudo package.

Checking and redefining the locale to C if the locale differs from en_US.UTF-8|C.UTF-8|C, for correct operation of the script with the server controller.

Checking that the sudo package is pre-installed if the user is not the root.

Checking the presence of the user or the required group in the sudoers file.

Checking permission to run the package manager via sudo.

Checking for no password prompt.

Implementation of using the user's home directory name to define the CUR_USER variable when whoami is missing.

Correction of descriptions and translations of errors: ServerUserNotInSudo and ServerPacketManagerError.

Checking the availability of the server user's home directory. Lack of such access leads to the incorrect information that the package manager is busy and Error 206 (ServerPacketManagerError). Or leads to the incorrect information that the server port already allocated  (Error 201)


client/server_scripts/check_user_in_sudo.sh
vanyasem, line 1: Shouldn't this script have a [shebang](https://en.wikipedia.org/wiki/Shebang_%28Unix%29) `#!/usr/bin/env bash` at the top?

client/server_scripts/prepare_host.sh
vanyasem, line 1: Shouldn't this script have a [shebang](https://en.wikipedia.org/wiki/Shebang_%28Unix%29) `#!/usr/bin/env bash` at the top?

client/server_scripts/check_user_in_sudo.sh
lunardunno: For what purposes?
This script is executed in sh.
Have you tried running it in sh?

vanyasem: You're right, a shebang won't help, as the script is executed line-by-line. My mistake

However, it uses the default user's shell instead of directly calling `sh`:

1) The script is ran in `ServerController::isUserInSudo`:
https://github.com/amnezia-vpn/amnezia-client/blob/157064763e0897e60fd9461bf0635ced38315a6d/client/core/controllers/serverController.cpp#L771

2) `ServerController::runScript` sends the command to an ssh client:
https://github.com/amnezia-vpn/amnezia-client/blob/157064763e0897e60fd9461bf0635ced38315a6d/client/core/controllers/serverController.cpp#L87

3) Finally, `Client::executeCommand` calls `libssh`:
https://github.com/amnezia-vpn/amnezia-client/blob/678bfffe492ed89911a9dcacfb125e82aa1ce1be/client/core/sshclient.cpp#L148

4) `ssh_channel_request_exec` _Run a shell command without an interactive shell._: https://api.libssh.org/master/group__libssh__channel.html#ga567d509183ade0a77190f390e2b5747d

You might get confused by the phrase `This is similar to 'sh -c command'.`

This implies the fact that a non-interactive shell is spawned, similar to how `sh -c` executes commands in a non-interactive shell. It does not mean that the commands are in fact executed in `sh`. Any shell has a non-interactive mode, and the default shell for the particular user is used to run the command.

There is no guarantee that the user's shell will be POSIX-compliant. For example, having `fish` or `pwsh` set as a login shell breaks the execution of the script. See https://github.com/amnezia-vpn/amnezia-client/issues/1201

This could be fixed by prefixing every line with `/bin/sh -c` or something similar.

At the very least, there could be a check that verifies that a supported (POSIX-compliant) shell is being used by the specified ssh user.

vanyasem: I actually made a small example to prove that your assumption that the code is ran in `/bin/sh` is wrong: https://gist.github.com/vanyasem/4c931f186c48da7b98b0f5518f1c6c64

`ssh_example` establishes an SSH connection via a private key, and then executes `ssh_channel_request_exec(channel, "echo $0")`. The output of the command is printed to console. `echo $0` returns the name of the currently running program, which in this case would be the name of the shell.

As you can see, changing the user's shell by calling `chsh` results in `$SHELL` being changed. If the commands were executed in `sh`, the value would always be `sh`

Screenshot 2025-03-04 at 12 04 52

Nethius (approved):

PR #1462, opened 2025-03-08
fix for sh

lunardunno:
Fix for servers where sh is used as default shell.


Nethius (approved):

vanyasem:
(Coming from https://github.com/amnezia-vpn/amnezia-client/issues/1201)

See, the command wasn't **always** executed in `sh` line-by-line as you initially said. It depends on the default shell specified for the user, which causes all kinds of compatibility issues.
I appreciate your work, thank you. Hopefully the shell compatibility issue will be fixed

PR #1500, opened 2025-03-29
fix: use iptables-nft if nftables is loaded

pgalonza:
It fix some problems when use nftables.


outspace:
please describe problems that you have with default iptables.

pgalonza:
@outspace 

The situations I have encountered are because I use a distribution that is not included in the recommended ones, which has nftables by default, but it may come in handy in future versions of amnezia.

1. When using iptables and iptables together, iptables rules are processed first. And iptables rules cannot be overridden.
2. If one of the chains in iptables is missing (because it exists in nftables), the rules will not be added.

PR #1557, opened 2025-04-30
OpenSUSE support

lunardunno:
Checking LOCK_FILE for zypper to support OpenSUSE


Nethius (approved):

Nethius (approved):

NHOrus:
In my experience, amnezia-client works on opensuse without this, with just the installation of systemd-resolved. But ok?

PR #1569, opened 2025-05-06
feature/premium v1 migration

Nethius:


client/ui/controllers/api/apiPremV1MigrationController.h
```diff
@@ -0,0 +1,50 @@
...
+    void errorOccurred(ErrorCode errorCode);
+
+    void showMigrationDrawer();
+    void migrationFinished();
+
+    void noSubscriptionToMigrate();
+
+private:
+    QSharedPointer<ServersModel> m_serversModel;
+    std::shared_ptr<Settings> m_settings;
```
aiamnezia, line 43: Why use different shared pointers here? 
Is it necessary?

client/ui/qml/Components/ApiPremV1MigrationDrawer.qml
```diff
@@ -0,0 +1,193 @@
...
+
+    expandedStateContent: Item {
+        implicitHeight: root.expandedHeight
+
+        ListViewType {
+            id: listView
+
+            anchors.fill: parent
+
+            model: 1 // fake model to force the ListView to be created without a model
```
aiamnezia, line 40: Why do we use ListView here if it doesn't have model at all?

aiamnezia (approved):

PR #1659, opened 2025-06-19
fixed issue when native connection format preserved after switching p…

Mitternacht822:


client/ui/qml/Pages2/PageShare.qml
```diff
@@ -458,6 +458,11 @@ PageType {
                         } else if (index === ContainerProps.containerFromString("amnezia-xray")) {
                             root.connectionTypesModel.push(xrayConnectionFormat)
                         }
+
+                        if (exportTypeSelector.currentIndex >= root.connectionTypesModel.length) {
```
Nethius: I think it would be better to move it to the handler() function, as it doesn't relate to this one

Nethius (approved):

PR #1665, opened 2025-06-23
fix: fixed desktop entry version for linux

### LINKED ISSUE #1664: Incorrect use of the `Version` field in .desktop file
**Describe the bug**

`Version` field is not for the application version, it is for the specification version (usually 1.0). Specifying an incorrect version may cause various problems in some DE.

```desktop
[Desktop Entry]
# ...

# The version of the desktop entry specification to which this file complies
Version=1.0

# ...
```

**Desktop (please complete the following information):**
 - OS: archlinux
 - Version: 4.8.7.2

Nethius:


lufog:
You can place this field right after `[Desktop Entry]` to emphasize that this is a specification version, not an application version. For example, chromium `.desktop` file.

```desktop
[Desktop Entry]
Version=1.0
Name=Chromium
# ...
```

PR #1803, opened 2025-08-20
fix: index assignment

MrMirDan:


vkamn:
fixed here https://github.com/amnezia-vpn/amnezia-client/pull/1817

PR #1848, opened 2025-09-09
fix: duplicate delete

vkamn:
…destructor


albexk:
Only this crash remains, but it’s also present in older versions of the app. It looks like Google doesn’t flag it.
```
09-09 17:33:14.791  3633  3683 F libc    : FORTIFY: pthread_mutex_lock called on a destroyed mutex (0x79335a7cd8)
09-09 17:33:14.791  3633  3683 F libc    : Fatal signal 6 (SIGABRT), code -1 (SI_QUEUE) in tid 3683 (hwuiTask0), pid 3633 (org.amnezia.vpn)
```

albexk (approved):

PR #1937, opened 2025-10-16
fix: checked format after changing protocol

MrMirDan:


Mitternacht822 (approved):

Mitternacht822 (approved):
Made a pair of additional fixes

PR #1945, opened 2025-10-22
Feat: file password protection

MrMirDan:
Encrypt and decrypt files with password using AES


Mitternacht822 (approved):

client/ui/controllers/systemController.cpp
```diff
@@ -89,6 +103,229 @@ bool SystemController::readFile(const QString &fileName, QString &data)
...
+                               EVP_sha256(), KEY_LEN, reinterpret_cast<unsigned char *>(outKey.data()));
+    if (!ok) {
+        qDebug() << opensslErrString();
+    }
+    return ok == 1;
+}
+
+static bool aesCrypt(const QByteArray &in, const QByteArray &key, const QByteArray &iv, QByteArray &out, bool encrypt)
+{
+    EVP_CIPHER_CTX *ctx = EVP_CIPHER_CTX_new();
```
ygurov: I would suggest to make it `unique_ptr` with custom destructor to handle throws properly and make the code more cleaner to maintain (no manual EVP_CIPHER_CTX_free will be required)


```suggestion
    std::unique_ptr<EVP_CIPHER_CTX, void (*)(EVP_CIPHER_CTX*)> ctx { EVP_CIPHER_CTX_new(), EVP_CIPHER_CTX_free };
```

client/ui/controllers/systemController.cpp
```diff
@@ -89,6 +103,229 @@ bool SystemController::readFile(const QString &fileName, QString &data)
...
+    int outlen1 = 0;
+    if (1
+        != EVP_CipherUpdate(ctx, reinterpret_cast<unsigned char *>(out.data()), &outlen1,
+                            reinterpret_cast<const unsigned char *>(in.constData()), in.size())) {
+        qDebug() << opensslErrString();
+        EVP_CIPHER_CTX_free(ctx);
+        return false;
+    }
+    int outlen2 = 0;
+    if (1 != EVP_CipherFinal_ex(ctx, reinterpret_cast<unsigned char *>(out.data()) + outlen1, &outlen2)) {
```
ygurov: OpenSSL states this in documentation:
```
The input length needs to be at least EVP_CIPHER_block_size() bytes long.
```
So we have to
```
out.resize(outlen1 + EVP_CIPHER_block_size(cipher));
```
prior this call. Or to check at least if the amount is enough for OpenSSL

MrMirDan: The buffer is already large enough, don't think additional resize() is required before EVP_CipherFinal_ex

PR #1972, opened 2025-11-02
feat: add system dns support

aiamnezia:


client/core/networkUtilities.cpp
```diff
@@ -13,6 +13,7 @@
     #include <QNetworkInterface>
     #include "qendian.h"
     #include <QSettings>
+    #pragma comment(lib, "iphlpapi.lib")
```
pokamest, line 16: not compatible with mingw

PR #2008, opened 2025-11-21
fix: Clear cache on start app

NickVs2015:
Add cache clear wher reinstall app


Mitternacht822 (commented):
May be it would be a better option to clean up cache not with every launch, but only when the app version changes. So it won't affect the startup time.

albexk (approved):

PR #2019, opened 2025-11-26
Fix: Application freeze on quit from tray

mystergaif:
Fixes an issue where the application would freeze when quitting from the system tray menu. Added a wait() call to the AmneziaApplication destructor to ensure the VPN connection thread terminates correctly.


vkamn:
This issue has already been fixed in this branch https://github.com/amnezia-vpn/amnezia-client/pull/1804

PR #2020, opened 2025-11-26
Feat/add msi installer

vkamn:


client/core/osSignalHandler.cpp
```diff
@@ -0,0 +1,67 @@
...
+OsSignalHandler::OsSignalHandler(QObject *parent) : QObject(parent)
+{
+}
+
+void OsSignalHandler::setup()
+{
+    if (signalHandler)
+        return;
+
+    signalHandler = new OsSignalHandler(QCoreApplication::instance());
```
ygurov: Currently, the allocation of a new class is redundant; there is no method which could be used by the instance of this class. Could be replaced with a plain bool variable

client/core/osSignalHandler.cpp
```diff
@@ -0,0 +1,67 @@
...
+#endif
+}
+
+static OsSignalHandler *signalHandler = nullptr;
+
+#if (defined(Q_OS_LINUX) && !defined(Q_OS_ANDROID)) || defined(Q_OS_MACX)
+void OsSignalHandler::handleSignal(int)
+{
+    if (QCoreApplication::instance()) {
+        QMetaObject::invokeMethod(QCoreApplication::instance(), "quit", Qt::QueuedConnection);
```
ygurov: `signal` or `sigaction` calls are asynchronous and support only a specific subset of standard functions.
POSIX declares those functions as async-signal-safe functions and has a full list here:
https://pubs.opengroup.org/onlinepubs/9699919799/functions/V2_chap02.html#tag_15_04

Basically, there is no new thread or memory allocation supported; the signal must be just delivered to the application.
I would suggest to consider using `sigwait` instead of asynchronous handling or do something like this instead
https://doc.qt.io/qt-6/unix-signals.html

client/core/osSignalHandler.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+void OsSignalHandler::setup()
+{
+    if (signalHandler)
+        return;
+
+    signalHandler = new OsSignalHandler(QCoreApplication::instance());
+
+#if (defined(Q_OS_LINUX) && !defined(Q_OS_ANDROID)) || defined(Q_OS_MACX)
+    signal(SIGINT, OsSignalHandler::handleSignal);
```
ygurov: `signal` is an old unsafe function that is not recommended to be used by POSIX.
I would suggest to consider using `sigaction` or `sigwait`

ygurov (changes requested):

PR #2210, opened 2026-02-12
fix(macos): fix permission error 1200 (QFile) in installer

desspperate:
The macOS installer (.pkg) creates the `/Library/Application Support/AmneziaVPN` directory with root ownership. When the user runs the app, it fails to write config files, causing "ErrorCode 1200: QFile error".

This commit adds a chown command to the postinstall script in `deploy/build_macos.sh` to transfer ownership of this directory to the current console user.


desspperate:
Any updates?

desspperate:
Any updates?

PR #2241, opened 2026-02-19
fix(Ubuntu 24.04): DNS resolution fails when VPN connects due to systemd-resolved restart

pavelzak:
flushDns() was restarting systemd-resolved via systemctl restart, which killed the D-Bus connection while async DNS configuration
   calls (SetLinkDNS, SetLinkDefaultRoute, Domains query) from DnsUtilsLinux were still in-flight. This caused all D-Bus calls to
  fail with NoReply, leaving DNS unconfigured on the VPN interface.

  - Use resolvectl flush-caches instead of restarting systemd-resolved, with fallback to restart for older distros
  - Add D-Bus error detail logging (error name + message) to dnsCallCompleted and dnsDomainsReceived
  - Add QDBusInterface validity check at construction time
  - Add errno details to ifindex resolution failure
  - Fix double-space typo in "DNS  domains" log message

### DIFF
--- a/client/daemon/daemonlocalserver.cpp
+++ b/client/daemon/daemonlocalserver.cpp
@@ -42,7 +42,8 @@ bool DaemonLocalServer::initialize() {
   }
 
   if (!m_server.listen(path)) {
-    logger.error() << "Failed to listen the daemon path";
+    logger.error() << "Failed to listen the daemon path:"
+                   << m_server.errorString();
     return false;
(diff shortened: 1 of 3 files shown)


pavelzak:
Hey, this is just some bug which I have encountered myself on my Ubuntu and fixed. Main fix is flush-caches the rest is the logging I have added to find the problem.

Issue #1591, opened 2025-05-16
Custom DNS settings ignored


agentsas:
**Describe the bug**

**To Reproduce**
1. Go to 'Settings'
2. Click on 'Connections'
3. Click on 'DNS Servers'
4. Set 'Primary DNS' to 8.8.8.8
5. Clear 'Secondary DNS'
6. Save settings
7. Connect to VPN
8. Open AmneziaVPN tunel status details
9. Value of 'IPv4 DNS server' is 1.1.1.1 instead of provided in configuration

I expect connection to use provided DNS settings

**Log files**
[AmneziaVPN.log](https://github.com/user-attachments/files/20241947/AmneziaVPN.log)

**Screenshots**
<img width="776" alt="Image" src="https://github.com/user-attachments/assets/f345f259-5fbf-4f8b-a7c2-01138375f2e9" />

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Version 24H2

**Server (please complete the following information):**
 - OS: Ubuntu 24.04

bqio:
Resolved?

qumuq-til:
For those struggling with the mystery of ignored (or wrong, not present in configuration) DNS servers, what worked for me was removing all DNS entries from the Amnezia client, then creating new users when sharing a connection. New users for the same connection didn't have those DNS servers and I could freely specify my own. It seems that when Amnezia creates a new user, it also writes DNS servers into the same connection file and then the options you specify manually in global Amnezia DNS settings just get ignored. If your connection link/file doesn't have any DNS servers then global options work.

woldemar89:
@qumuq-til 
Many thanks! Works for me!

I tried to change DNS servers and had same issue described.

I was using a settings file I created earlier and it looks to override app DNS settings..

Now if I set DNS addresses first, then Share(create) a file, then use it to Connect, - it finally works and I see correct DNS IPs in Windows connection!

Leaving DNS settings empty, then Share(create) a file, then use it works too! (don't forget to restore\fill app DNS settings back, otherwise it does not connect). 

p.s. Amnezia Support replied very odd about it: they recommended to install AmneziaDNS or set DNS in a browser..

PR #2273, opened 2026-02-24
Fix custom DNS for non-Admin profiles + preserve DNS order in Android WireGuard

ilsme:
## Problem
  For non-API profiles (`config_version == 0`), Custom DNS from app settings could be ignored.
  This was especially visible for shared/imported self-hosted profiles, where DNS embedded in the profile overrode GUI Custom DNS.

Additionally, Android WireGuard could apply DNS in the wrong primary/secondary order.

## Root cause
- DNS selection for non-API profiles did not consistently prioritize app DNS settings.
- `ServersModel::getDnsPair()` was not strictly tied to the requested server index.
- Android WG config generation did not preserve DNS order from input.

## Solution
- Updated `ServersModel::getDnsPair(serverIndex)` to always use the requested `serverIndex`.
- For non-API profiles (`config_version == 0`), prefer app Custom DNS (`primaryDns`/`secondaryDns`).
- Kept API-config behavior unchanged (server-provided DNS is still used there).
- Preserved primary/secondary DNS order in Android WireGuard config generation.

## Result
- Custom DNS now works for shared/imported non-API profiles (including self-hosted flows).
- Android WireGuard keeps DNS order exactly as configured.

## Related issues
- #1591

### DIFF
--- a/client/android/protocolApi/src/main/kotlin/ProtocolConfig.kt
+++ b/client/android/protocolApi/src/main/kotlin/ProtocolConfig.kt
@@ -40,7 +40,8 @@ open class ProtocolConfig protected constructor(
 
     open class Builder(blockingMode: Boolean) {
         internal val addresses: MutableSet<InetNetwork> = hashSetOf()
-        internal val dnsServers: MutableSet<InetAddress> = hashSetOf()
+        // Preserve insertion order: Android treats the first DNS as primary.
+        internal val dnsServers: MutableSet<InetAddress> = linkedSetOf()
         internal val routes: MutableSet<Route> = mutableSetOf()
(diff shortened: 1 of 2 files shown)


vkamn:
This is not a bug it is intentional behavior, DNS can only be changed for the admin config. It’s just poorly implemented from a UX/UI perspective

ilsme:
Yes, that's what I thought, but I think it would be better to rework UX/UI and give the ability for client to choose whether to use config embedded DNS or Custom DNS. 
Atm it's really confusing which DNS settings are in use and how AmneziaDNS is involved if availible(as I understood it's used as a fallback only and it's not what I thought before I dove into the code).

If it's no an option then I'll close this PR.

PR #2322, opened 2026-03-07
Optimize Dockerfiles: reduce image size and layer count across all services

seventhsite:
docker: optimize Dockerfile to reduce image size and layer count

- Merge all RUN instructions into a single layer
- Replace echo -e with printf and heredocs for portability
- Add --update flag to apk to avoid separate upgrade step
- Clean up apk cache, temp dirs and terminfo after install
- Remove redundant CMD [""] and unnecessary tee/sed usage


vkamn:
https://github.com/amnezia-vpn/amnezia-client/pull/2324 duplicate

PR #2370, opened 2026-03-18
Feat: Add MtProxy (Telegram)

yp-amnezia:


BLaDZer:
this is useless because mtproxy is not a VPN, but amnezia-vpn mean to use VPN services...

client/server_scripts/mtproxy/start.sh
```diff
@@ -0,0 +1,73 @@
...
+
+# Build domain argument for FakeTLS mode
+DOMAIN_ARG=""
+if [ "$MTPROXY_TRANSPORT_MODE" = "faketls" ] && [ -n "$MTPROXY_TLS_DOMAIN" ]; then
+    DOMAIN_ARG="--domain $MTPROXY_TLS_DOMAIN"
+fi
+
+WORKERS=$MTPROXY_WORKERS
+STATS_PORT=2398
+LISTEN_PORT=$MTPROXY_PORT
```
aiamnezia, line 31: In run_container.sh, Docker maps -p $MTPROXY_PORT:443/tcp (host port → container port 443). But here, the line causes mtproto-proxy to listen on $MTPROXY_PORT inside the container. When the user configures any port other than the default 443, the proxy listens on the wrong port inside the container.

client/server_scripts/mtproxy/run_container.sh
```diff
@@ -0,0 +1,9 @@
+# Run container
+sudo docker run -d \
+  --log-driver none \
+  --restart always \
+  -p $MTPROXY_PORT:443/tcp \
```
aiamnezia:
```suggestion
  -p $MTPROXY_PORT:$MTPROXY_PORT/tcp \
```

client/server_scripts/mtproxy/Dockerfile
aiamnezia, line 1: Can we make an image with the proxy and upload it on Docker Hub as for awg for example? Thus we could maintain the image independently of the client

xRayDev:
> this is useless because mtproxy is not a VPN, but amnezia-vpn mean to use VPN services...

Its not VPN, but this service like socsk5 proxy

aiamnezia (changes requested):

vkamn (approved):

PR #2415, opened 2026-03-26
fix: add linux reconnection

NickVs2015:
fix: reconnection on linux

### DIFF
--- a/client/mozilla/localsocketcontroller.cpp
+++ b/client/mozilla/localsocketcontroller.cpp
@@ -27,7 +27,7 @@
 #include "protocols/protocols_defs.h"
 
 // How many times do we try to reconnect.
-constexpr int MAX_CONNECTION_RETRY = 70;
+constexpr int MAX_CONNECTION_RETRY = 10;
 
 // How long do we wait between one try and the next one.
 constexpr int CONNECTION_RETRY_TIMER_MSEC = 500;
--- a/client/mozilla/localsocketcontroller.cpp
+++ b/client/mozilla/localsocketcontroller.cpp
@@ -27,7 +27,7 @@
 #include "protocols/protocols_defs.h"
 
 // How many times do we try to reconnect.
-constexpr int MAX_CONNECTION_RETRY = 10;
+constexpr int MAX_CONNECTION_RETRY = 70;
 
 // How long do we wait between one try and the next one.
 constexpr int CONNECTION_RETRY_TIMER_MSEC = 500;
--- a/client/platforms/linux/daemon/dnsutilslinux.cpp
+++ b/client/platforms/linux/daemon/dnsutilslinux.cpp
@@ -7,6 +7,7 @@
 #include <net/if.h>
 
 #include <QDBusVariant>
+#include <QTimer>
 #include <QtDBus/QtDBus>
 
 #include "leakdetector.h"
@@ -58,6 +59,7 @@ bool DnsUtilsLinux::updateResolvers(const QString& ifname,
     return false;
   }
 
+  m_resolvers = resolvers;
   setLinkDNS(m_ifindex, resolvers);
   setLinkDefaultRoute(m_ifindex, true);
   updateLinkDomains();
@@ -90,7 +92,7 @@ bool DnsUtilsLinux::restoreResolvers() {
 void DnsUtilsLinux::dnsCallCompleted(QDBusPendingCallWatcher* call) {
   QDBusPendingReply<> reply = *call;
   if (reply.isError()) {
-    logger.error() << "Error received from the DBus service";
+    logger.debug() << "DBus call failed (may be transient after systemd-resolved restart)";
   }
   delete call;
 }
@@ -174,11 +176,25 @@ void DnsUtilsLinux::updateLinkDomains() {
 
 void DnsUtilsLinux::dnsDomainsReceived(QDBusPendingCallWatcher* call) {
   QDBusPendingReply<QVariant> reply = *call;
+  delete call;
   if (reply.isError()) {
-    logger.error() << "Error retrieving the DNS  domains from the DBus service";
-    delete call;
+    // systemd-resolved may still be starting up after a restart — retry a few times
+    if (m_domainRetries++ < 5) {
+      logger.debug() << "systemd-resolved not ready yet, retrying DNS setup ("
+                     << m_domainRetries << "/5)";
+      // Re-apply DNS servers and default route in case systemd-resolved lost them on restart
+      if (m_ifindex > 0 && !m_resolvers.isEmpty()) {
+        setLinkDNS(m_ifindex, m_resolvers);
+        setLinkDefaultRoute(m_ifindex, true);
+      }
+      QTimer::singleShot(500, this, &DnsUtilsLinux::updateLinkDomains);
+    } else {
+      logger.warning() << "Failed to configure DNS after 5 retries";
+      m_domainRetries = 0;
+    }
     return;
   }
+  m_domainRetries = 0;
 
   /* Update the state of the DNS domains */
   m_linkDomains.clear();
@@ -206,7 +222,6 @@ void DnsUtilsLinux::dnsDomainsReceived(QDBusPendingCallWatcher* call) {
   /* Add a root search domain for the new interface. */
   QList<DnsLinkDomain> newlist = {root};
   setLinkDomains(m_ifindex, newlist);
-  delete call;
 }
 
 static DnsMetatypeRegistrationProxy s_dnsMetatypeProxy;
--- a/client/platforms/linux/daemon/dnsutilslinux.cpp
+++ b/client/platforms/linux/daemon/dnsutilslinux.cpp
@@ -7,7 +7,6 @@
 #include <net/if.h>
 
 #include <QDBusVariant>
-#include <QTimer>
 #include <QtDBus/QtDBus>
 
 #include "leakdetector.h"
@@ -61,7 +60,6 @@ bool DnsUtilsLinux::updateResolvers(const QString& ifname,
 
   setLinkDNS(m_ifindex, resolvers);
   setLinkDefaultRoute(m_ifindex, true);
-  setLinkDomains(m_ifindex, {DnsLinkDomain(".", true)});
   updateLinkDomains();
   return true;
 }
@@ -92,7 +90,7 @@ bool DnsUtilsLinux::restoreResolvers() {
 void DnsUtilsLinux::dnsCallCompleted(QDBusPendingCallWatcher* call) {
   QDBusPendingReply<> reply = *call;
   if (reply.isError()) {
-    logger.debug() << "DBus call failed (may be transient after systemd-resolved restart)";
+    logger.error() << "Error received from the DBus service";
   }
   delete call;
 }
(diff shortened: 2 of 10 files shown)


client/mozilla/localsocketcontroller.cpp
```diff
@@ -27,7 +27,7 @@
 #include "protocols/protocols_defs.h"
 
 // How many times do we try to reconnect.
-constexpr int MAX_CONNECTION_RETRY = 10;
+constexpr int MAX_CONNECTION_RETRY = 70;
```
ygurov: 70 retries seems too much. Why do we change that?
CONNECTION_RETRY_TIMER_MSEC * MAX_CONNECTION_RETRY => 70 * 500msec => 35sec is 1/4 of the whole session

client/platforms/linux/daemon/dnsutilslinux.cpp
```diff
@@ -174,11 +176,25 @@ void DnsUtilsLinux::updateLinkDomains() {
 
 void DnsUtilsLinux::dnsDomainsReceived(QDBusPendingCallWatcher* call) {
   QDBusPendingReply<QVariant> reply = *call;
+  delete call;
```
ygurov: Use of `delete` for Qt objects could be UB in some cases. Use `deleteLater()` to be sure it's done in event loop

client/platforms/linux/linuxnetworkwatcherworker.cpp
```diff
@@ -199,10 +208,19 @@ void LinuxNetworkWatcherWorker::checkDevices() {
...
+
   if (state == NM_STATE_ASLEEP) {
     emit wakeup();
+  } else if (state >= NM_STATE_CONNECTED_SITE && m_previousNMState < NM_STATE_CONNECTED_SITE) {
+    // Delay the reconnect so the kernel routing table (and default gateway)
+    // are fully restored before the VPN exclusion route is recalculated.
+    // Without this delay, getGatewayAndIface() may return an empty gateway
+    // immediately after network recovery, causing the server exclusion route
+    // to be installed with gateway 0.0.0.0 and breaking WG handshakes.
+    QTimer::singleShot(3000, this, [this]() { emit networkChanged(); });
```
ygurov: Using constant delay here is a bad idea since we have a bunch of different setups based on Linux. Some of them could have a small routing table, and this is postponing the actual reconnection event, while the others could not to make a full reinitialization in this timing.
What if we check here if the routes are established?

client/platforms/linux/linuxnetworkwatcherworker.cpp
```diff
@@ -199,10 +208,19 @@ void LinuxNetworkWatcherWorker::checkDevices() {
 
 void LinuxNetworkWatcherWorker::NMStateChanged(quint32 state)
 {
+  logger.debug() << "NMStateChanged " << state;
+
   if (state == NM_STATE_ASLEEP) {
```
ygurov: NM_STATE_DISABLED should be handled here as well

client/platforms/linux/daemon/dnsutilslinux.cpp
```diff
@@ -174,11 +176,25 @@ void DnsUtilsLinux::updateLinkDomains() {
...
   if (reply.isError()) {
-    logger.error() << "Error retrieving the DNS  domains from the DBus service";
-    delete call;
+    // systemd-resolved may still be starting up after a restart — retry a few times
+    if (m_domainRetries++ < 5) {
+      logger.debug() << "systemd-resolved not ready yet, retrying DNS setup ("
+                     << m_domainRetries << "/5)";
+      // Re-apply DNS servers and default route in case systemd-resolved lost them on restart
+      if (m_ifindex > 0 && !m_resolvers.isEmpty()) {
```
ygurov: The only place `dnsDomainsReceived` is getting called is by this chain:
```
updateResolvers -> updateLinkDomains -> dnsDomainsReceived
```
And `updateResolvers` does the same things you are trying to address here:
```
setLinkDNS(m_ifindex, resolvers);
setLinkDefaultRoute(m_ifindex, true);
```

That's a big hack and a huge bugprone thing that must be removed

client/platforms/linux/linuxnetworkwatcherworker.cpp
```diff
@@ -122,6 +122,15 @@ void LinuxNetworkWatcherWorker::initialize() {
         SLOT(propertyChanged(QString, QVariantMap, QStringList)));
   }
 
+  // Seed m_previousNMState with the current NM state so that a transient
+  // drop from CONNECTED_GLOBAL (70) to CONNECTED_SITE (60) caused by the
```
ygurov: The comment displays not the things you are trying to do here and later in the code. Should be omitted at all, the code is crystal clear

client/platforms/linux/daemon/linuxfirewall.cpp
```diff
@@ -289,6 +289,7 @@ void LinuxFirewall::install()
     installAnchor(Both, QStringLiteral("100.blockAll"), {
                                                             QStringLiteral("-j REJECT"),
                                                         });
+    installAnchor(Both, QStringLiteral("400.allowPIA"), {});
```
ygurov: Nice, but should be addressed in the other PR I think

client/platforms/linux/daemon/linuxfirewall.cpp
```diff
@@ -495,13 +496,23 @@ int LinuxFirewall::execute(const QString &command, bool ignoreErrors)
...
-    if (!err.isEmpty())
+    if (!err.isEmpty() && !ignoreErrors)
         logger.warning() << err;
     return exitCode;
 }
 
 void LinuxFirewall::setupTrafficSplitting()
 {
+    // Register the routing table name so the ip tool can resolve it by name
+    execute(QStringLiteral("grep -q '%1' /etc/iproute2/rt_tables || printf '200\\t%1\\n' >> /etc/iproute2/rt_tables").arg(kRtableName));
```
ygurov: What is the purpose of those? I do not think touching firewall/killswitch is a good idea in boundaries of this PR

ygurov (changes requested):

client/platforms/linux/linuxnetworkwatcherworker.cpp
NickVs2015: switch to use NetworkUtilities::getGatewayAndIface() like in other places

client/mozilla/localsocketcontroller.cpp
NickVs2015: QDBusInterface constructor in DnsUtilsLinux blocks the daemon's main thread for up to 25 seconds waiting for systemd-resolved to appear on D-Bus. That is why MAX_CONNECTION_RETRY =10  not enought

client/platforms/linux/daemon/linuxfirewall.cpp
NickVs2015: move to separate PR https://github.com/amnezia-vpn/amnezia-client/pull/2433

client/platforms/linux/daemon/linuxfirewall.cpp
NickVs2015: move to separate PR https://github.com/amnezia-vpn/amnezia-client/pull/2433

client/mozilla/localsocketcontroller.cpp
NickVs2015: restore to default

client/platforms/linux/daemon/wireguardutilslinux.cpp
```diff
@@ -357,13 +361,25 @@ bool WireguardUtilsLinux::addExclusionRoute(const IPAddress& prefix) {
     if (!m_rtmonitor) {
         return false;
     }
+    if (++m_exclusionRefCount[prefix] > 1) {
```
ygurov: Why do we make those changes? I do not see any reason and find this mechanism fragile to keep in the code

client/platforms/linux/linuxnetworkwatcherworker.cpp
ygurov: Did not really get the point. I think we should handle both legacy `NM_STATE_ASLEEP` and the newer one `NM_STATE_DISABLED` for the `wakeup` case

client/platforms/linux/linuxnetworkwatcherworker.cpp
```diff
@@ -199,10 +207,44 @@ void LinuxNetworkWatcherWorker::checkDevices() {
...
+  m_previousNMState = state;
+}
+
+void LinuxNetworkWatcherWorker::checkGatewayAndEmit(int generation, int count) {
+  if (m_pollGeneration.load() != generation) {
+    return;
+  }
+
+  ++count;
+  QString gateway = NetworkUtilities::getGatewayAndIface().first;
```
ygurov: When we have VPN connected, the default gateway and interface is the TUN interface. It should not be used during VPN connection, otherwise it will return TUN interface address and interface

ygurov (changes requested):

client/platforms/linux/linuxnetworkwatcherworker.cpp
```diff
@@ -199,10 +207,13 @@ void LinuxNetworkWatcherWorker::checkDevices() {
...
-  if (state == NM_STATE_ASLEEP) {
-    emit wakeup();
-  }
+    logger.debug() << "NMStateChanged " << state;
 
-  logger.debug() << "NMStateChanged " << state;
-}
+    if (state == NM_STATE_ASLEEP || state == NM_STATE_DISABLED) {
+        emit wakeup();
+    } else if (state >= NM_STATE_CONNECTED_SITE && m_previousNMState < NM_STATE_CONNECTED_SITE) {
```
ygurov: Is there any point of tracking the previous state?
Theoretically, there should be an event generated every time the network changes its status. And if it's happening during the reconnection/connection process - it should not be handled by the client

client/platforms/linux/daemon/dnsutilslinux.cpp
```diff
@@ -173,12 +175,21 @@ void DnsUtilsLinux::updateLinkDomains() {
...
 
 void DnsUtilsLinux::dnsDomainsReceived(QDBusPendingCallWatcher* call) {
-  QDBusPendingReply<QVariant> reply = *call;
+    QDBusPendingReply<QVariant> reply = *call;
+    call->deleteLater();
   if (reply.isError()) {
-    logger.error() << "Error retrieving the DNS  domains from the DBus service";
-    delete call;
+    // systemd-resolved may still be starting up after a restart — retry a few times
+    if (m_domainRetries++ < 5) {
```
ygurov: Let's kick off a different PR with these changes

client/platforms/linux/daemon/dnsutilslinux.cpp
```diff
@@ -60,6 +61,7 @@ bool DnsUtilsLinux::updateResolvers(const QString& ifname,
 
   setLinkDNS(m_ifindex, resolvers);
   setLinkDefaultRoute(m_ifindex, true);
+  setLinkDomains(m_ifindex, {DnsLinkDomain(".", true)});
```
ygurov: Same here, let's do it in a different PR

ygurov (changes requested):

ygurov (approved):

PR #2453, opened 2026-04-08
Require SOCKS5 auth on local Xray inbound (fixes #2452)

### LINKED ISSUE #2452: [Android] Per-app split tunneling bypass via unauthenticated local SOCKS5 listener on 127.0.0.1:10808
**Summary**

When AmneziaVPN is active on Android (X-ray profile), it exposes a SOCKS5 listener on 127.0.0.1:10808 that accepts unauthenticated connections. Because Android loopback is shared across apps, another application can connect to this listener. As a result, an app excluded from VPN routing via per-app split tunneling can still send traffic through the VPN by using the local SOCKS5 proxy. This behavior may unintentionally weaken the isolation guarantees of split tunneling and could increase the detectability of VPN usage in environments where app-level network monitoring is employed.

**Environment** 
- **App version**: 4.8.14.5
- **Android version**: 15
- **Device**: Samsung Galaxy S25
- **Protocol used**: Xray
- **Split tunneling**: Enabled (Termux excluded from VPN)

**To Reproduce**
1. Install AmneziaVPN and configure a working connection.
2. Enable **Per-app VPN** (split tunneling) and **exclude Termux** (or any test app with `INTERNET` permission).
3. Connect to the VPN.
4. In Termux, run:
   ```bash
   # Verify that direct traffic is not using the VPN
   curl -s https://ifconfig.me/ip
   # Expected: your real public IP
   
   # Check whether the localhost proxy is reachable
   nc -vz -w1 127.0.0.1 10808
   # Observed: connection succeeds only while VPN is active
   
   # Send traffic through the local SOCKS5 proxy from excluded app
   curl -s --socks5-hostname 127.0.0.1:10808 --connect-timeout 10 https://ifconfig.me/ip
   # Observed: VPN exit IP

**Observed behavior**
- Port `127.0.0.1:10808` is open **only when VPN is active**.
- The listener accepts SOCKS5 connections without authentication.
- An app excluded from VPN routing can still reach the VPN exit by using this localhost proxy.

**Expected Behavior**
Applications excluded from VPN routing should not be able to send traffic through the VPN tunnel via a localhost proxy, unless this behavior is explicitly documented and user-controlled. If a local listener is required for internal purposes, access should be restricted to the app's own components.

**Security Impact**
If the proxy accepts unauthenticated connections and forwards traffic through the VPN tunnel:
- User-configured per-app routing policies can be bypassed by any app with INTERNET permission.
- Traffic that the user explicitly intended to route directly may leak through the VPN (or vice versa).
- The presence of an open, unauthenticated localhost proxy may serve as a detectable signal of VPN usage to monitoring applications.

**Operational relevance**
In environments where applications or platforms monitor network behavior for signs of VPN usage, this issue may increase the detectability of VPN use. Because an app excluded from VPN routing can still access the VPN tunnel via localhost, it can both infer that the VPN is active and send probe traffic through the tunnel. This may also aid enumeration of VPN egress IPs and contribute to blocking of VPN endpoints. This consideration is particularly relevant given recent reports of app-level VPN detection guidelines being distributed to IT platforms [1][2].

**Suggested Mitigations**
- Require SOCKS5 authentication (username/password) for the localhost listener, if the feature is intended for external use.
- If the listener is only needed for internal IPC, consider using a private communication channel (e.g., Unix domain socket with restrictive permissions, or Android Binder with UID verification) instead of a shared TCP port on loopback.
- Document the behavior and its implications in the app's settings or security notes, so users can make informed decisions about split tunneling usage.


**References**
1. **Technical analysis of localhost proxy bypass in mobile VPN clients** (Russian):  
   https://habr.com/ru/articles/1020080/
   
2. **Community discussion: MinDigital guidelines for VPN detection** (Russian):  
   https://habr.com/ru/news/1019646/  
   *Summary: Russian Ministry of Digital Development issued guidelines for IT platforms to detect VPN usage via app-level network monitoring (ConnectivityManager/NetworkCapabilities APIs on Android). The report highlights split tunneling as a known challenge for detection, making proper isolation of per-app routing critically important.*

Gerk17:
## Summary

Closes #2452.

The local SOCKS5 listener that bridges `tun2socks` and the Xray core was
unauthenticated and bound to `127.0.0.1:10808`. On Android the loopback
interface is shared across UIDs, so any third-party app with `INTERNET`
permission could connect to it and use it as an open proxy, bypassing
per-app split tunneling and sending traffic through the VPN tunnel even
when the user had explicitly excluded the app.

This PR requires SOCKS5 username/password authentication on that
inbound. Credentials are generated per process via
`QRandomGenerator::system()`, kept only in memory, and re-rolled on
every process start. The tun2socks consumer on each platform reads the
credentials and authenticates, so the VPN keeps working end-to-end
while external apps can no longer use the proxy.

## Changes

**Serialization (`client/core/serialization/`)**
- `inbound.cpp`: `GenerateInboundEntry()` now emits
  `settings.auth = "password"` and `settings.accounts = [{user, pass}]`
  with random 24-char tokens generated by `QRandomGenerator::system()`.
  Adds `GetInboundCredentials()` returning the same pair for in-process
  consumers via a Meyers singleton.
- `serialization.h`: declares `InboundCredentials` and
  `GetInboundCredentials()` next to the existing `GenerateInboundEntry()`.

**Desktop (`client/protocols/xrayprotocol.cpp`)**
- The tun2socks process is now launched with
  `socks5://user:pass@127.0.0.1:10808` instead of the unauthenticated
  URL. The credentials are read from `GetInboundCredentials()` so they
  always match what `GenerateInboundEntry()` put into the Xray config.

**Android (`client/android/xray/src/main/kotlin/`)**
- `XrayConfig.kt`: adds `socksUser` / `socksPass` fields, builder
  setters and constructor wiring, mirroring the existing `socksPort`.
- `Xray.kt`: parses
  `inbounds[0].settings.accounts[0]` from the Xray JSON config and
  populates the new fields. `runTun2Socks()` builds a
  `socks5://user:pass@127.0.0.1:port` URL for `Tun2SocksConfig.proxy`.
  Uses `optJSONObject` / `optJSONArray` so older configs without an
  `accounts` block degrade gracefully (no auth → behaves as before).

**iOS (`client/platforms/ios/PacketTunnelProvider+Xray.swift`)**
- Reads the same `inbounds[0].settings.accounts[0]` from the JSON
  config and threads `username` / `password` into
  `setupAndRunTun2socks(...)`. The hev-socks5-tunnel YAML config now
  includes `username:` / `password:` lines under `socks5:` when
  credentials are present.

## Security properties

- Credentials are generated by a CSPRNG (`QRandomGenerator::system()`,
  which maps to `/dev/urandom` on Linux/Android, `SecRandomCopyBytes`
  on Apple, `BCryptGenRandom` on Windows).
- They live only in process memory, are never written to disk, logs,
  or notifications, and are never serialized into shareable Xray
  config exports.
- They are re-rolled on every process start, so a VPN reconnect after
  an app kill yields a fresh pair.
- After the patch, an external app connecting to `127.0.0.1:10808`
  receives a SOCKS5 handshake that requires method `0x02`
  (username/password) and cannot proceed without the secret. The
  per-app split-tunneling bypass from #2452 is closed.

## What this PR does *not* do

It does not hide the fact that *something* is listening on
`127.0.0.1:10808`. An attacker app can still probe the port and infer
"AmneziaVPN with Xray is active". Fully addressing the
detectability concern from the second half of #2452 would require
moving the IPC off loopback TCP entirely (e.g. to a Unix domain
socket inside the app's private files dir), which in turn requires
adding Unix socket support upstream in `hev-socks5-tunnel`. That
should be tracked as a follow-up; this PR is the minimal fix that
closes the bypass without new dependencies.

## Testing

Locally on Linux desktop with Qt 6.6.2:
- Full client builds cleanly (`235/235`, both `AmneziaVPN` and
  `AmneziaVPN-service` link successfully).
- A small standalone test calling `GenerateInboundEntry()` and
  `GetInboundCredentials()` confirms that the JSON contains
  `auth: "password"` and a populated `accounts` array, that the
  credentials returned by `GetInboundCredentials()` match the ones
  inside the JSON, and that they differ between process invocations.

The Android, iOS, Windows and macOS changes are mechanical and rely
on this repository's CI for build verification. Reviewers familiar
with each platform should sanity-check the wiring; I'm happy to
iterate on anything that looks off.

### DIFF
--- a/client/android/xray/src/main/kotlin/Xray.kt
+++ b/client/android/xray/src/main/kotlin/Xray.kt
@@ -99,6 +99,13 @@ class Xray : Protocol() {
 
             val socksConfig = xrayJsonConfig.getJSONArray("inbounds")[0] as JSONObject
             socksConfig.getInt("port").let { setSocksPort(it) }
+            socksConfig.optJSONObject("settings")
+                ?.optJSONArray("accounts")
+                ?.optJSONObject(0)
+                ?.let {
+                    setSocksUser(it.optString("user"))
+                    setSocksPass(it.optString("pass"))
+                }
 
             configSplitTunneling(config)
             configAppSplitTunneling(config)
@@ -162,9 +169,12 @@ class Xray : Protocol() {
     }
 
     private fun runTun2Socks(config: XrayConfig, fd: Int) {
+        val auth = if (config.socksUser.isNotEmpty() && config.socksPass.isNotEmpty())
+                       "${config.socksUser}:${config.socksPass}@"
+                   else ""
         val tun2SocksConfig = Tun2SocksConfig().apply {
             mtu = config.mtu.toLong()
-            proxy = "socks5://127.0.0.1:${config.socksPort}"
+            proxy = "socks5://${auth}127.0.0.1:${config.socksPort}"
             device = "fd://$fd"
             logLevel = "warn"
         }
--- a/client/android/xray/src/main/kotlin/XrayConfig.kt
+++ b/client/android/xray/src/main/kotlin/XrayConfig.kt
@@ -9,17 +9,23 @@ private const val XRAY_DEFAULT_MAX_MEMORY: Long = 50 shl 20 // 50 MB
 class XrayConfig protected constructor(
     protocolConfigBuilder: ProtocolConfig.Builder,
     val socksPort: Int,
+    val socksUser: String,
+    val socksPass: String,
     val maxMemory: Long,
 ) : ProtocolConfig(protocolConfigBuilder) {
 
     protected constructor(builder: Builder) : this(
         builder,
         builder.socksPort,
+        builder.socksUser,
+        builder.socksPass,
         builder.maxMemory
     )
 
     class Builder : ProtocolConfig.Builder(false) {
         internal var socksPort: Int = 0
+        internal var socksUser: String = ""
+        internal var socksPass: String = ""
             private set
 
         internal var maxMemory: Long = XRAY_DEFAULT_MAX_MEMORY
@@ -29,6 +35,10 @@ class XrayConfig protected constructor(
 
         fun setSocksPort(port: Int) = apply { socksPort = port }
 
+
+        fun setSocksUser(user: String) = apply { socksUser = user }
+
+        fun setSocksPass(pass: String) = apply { socksPass = pass }
         fun setMaxMemory(maxMemory: Long) = apply { this.maxMemory = maxMemory }
 
         override fun build(): XrayConfig = configBuild().run { XrayConfig(this@Builder) }
--- a/client/core/serialization/inbound.cpp
+++ b/client/core/serialization/inbound.cpp
@@ -1,9 +1,12 @@
-#include <QString>
+#include <QJsonArray>
 #include <QJsonObject>
 #include <QList>
+#include <QRandomGenerator>
+#include <QString>
+
 #include "3rd/QJsonStruct/QJsonIO.hpp"
-#include "transfer.h"
 #include "serialization.h"
+#include "transfer.h"
 
 namespace amnezia::serialization::inbounds
 {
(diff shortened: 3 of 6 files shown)


vkamn:
already fixed in another PR

PR #2456, opened 2026-04-08
fix: tun2socks auth settings

yp-amnezia:


client/android/xray/src/main/kotlin/Xray.kt
```diff
@@ -173,6 +188,35 @@ class Xray : Protocol() {
         }
     }
 
+    // Ensures SOCKS5 auth is present in inbounds[0].settings.
+    // Re-uses existing credentials if already configured; otherwise generates random ones.
+    private fun ensureInboundAuth(xrayConfig: JSONObject) {
+        val inbounds = xrayConfig.optJSONArray("inbounds") ?: return
+        if (inbounds.length() == 0) return
+
+        val inbound = inbounds.getJSONObject(0)
```
ygurov: Theoretically, `socks` inbound could be not the first one in the list. Let's iterate over the all elements to find the right one with socks

client/core/serialization/inbound.cpp
```diff
@@ -14,25 +16,96 @@ namespace amnezia::serialization::inbounds
...
+    creds.username = account.value("user").toString();
+    creds.password = account.value("pass").toString();
+    return creds;
+}
+
+InboundCredentials EnsureInboundAuth(QJsonObject &xrayConfig)
+{
+    QJsonArray inbounds = xrayConfig.value("inbounds").toArray();
+    if (inbounds.isEmpty())
+        return GetInboundCredentials(xrayConfig); // degenerate config, nothing to patch
```
ygurov: Sounds offensive 😄

client/core/serialization/inbound.cpp
```diff
@@ -14,25 +16,96 @@ namespace amnezia::serialization::inbounds
...
+
+    QJsonObject inbound = inbounds.first().toObject();
```
ygurov: The same thing as the one above, theoretically `socks` could be not the first one inbound provider here. Let's iterate over all elements to find the right one

client/android/xray/src/main/kotlin/Xray.kt
```diff
@@ -26,6 +28,15 @@ import org.json.JSONObject
 private const val TAG = "Xray"
 private const val LIBXRAY_TAG = "libXray"
 
+@Suppress("SwallowedException")
+private fun acquireFreeLocalPort(): Int {
+    return try {
+        ServerSocket(0, 1, InetAddress.getByName("127.0.0.1")).use { it.localPort }
+    } catch (_: Exception) {
+        (49152..65535).random()
```
ygurov: That's not the best idea I think.
Do we have any cases we cannot open any port?

client/core/serialization/inbound.cpp
```diff
@@ -27,6 +29,26 @@ const static QString listen = "127.0.0.1";
...
+    constexpr int maxPort = 65535;
+    return QRandomGenerator::system()->bounded(minPort, maxPort + 1);
+}
+
+// Ask the OS for a free TCP port on loopback (same stack as inbound "listen": "127.0.0.1").
+static int acquireFreeLocalPort()
+{
+    QTcpServer probe;
+    if (!probe.listen(QHostAddress(QStringLiteral("127.0.0.1")), 0)) {
+        int port = generateEphemeralPort();
```
ygurov: The same here. Do we have any troubles with `QTcpServer::listen`?
I think generating a random port is not the best idea

client/platforms/ios/PacketTunnelProvider+Xray.swift
```diff
@@ -21,6 +22,36 @@ extension Constants {
 }
 
 extension PacketTunnelProvider {
+    /// TCP port chosen by the OS on IPv6 loopback (::1), matching inbound listen address.
+    private func acquireFreeLocalPort() -> Int {
+        let fd = socket(AF_INET6, SOCK_STREAM, IPPROTO_TCP)
+        guard fd != -1 else { return Int.random(in: 49152...65535) }
```
ygurov: The same as above

client/core/serialization/inbound.cpp
```diff
@@ -14,25 +19,125 @@ namespace amnezia::serialization::inbounds
...
+            return static_cast<int>(probe.serverPort());
+    }
+    return std::nullopt;
+}
+
+// Generates a hex string of `byteCount` random bytes (URL-safe, no special chars).
+static QString generateRandomHex(int byteCount)
+{
+    QByteArray buf(byteCount, 0);
+    QRandomGenerator::system()->fillRange(reinterpret_cast<quint32 *>(buf.data()),
```
ygurov: Here we would have a leakage in case of using numbers non-dividable by 4. I know this is not the current case, however I think this should be addressed to cover any further potential problems.
I would recommend to make `buf` to the size of `byteCount + sizeof(quint32) - 1) / sizeof(quint32)` and process it then

client/protocols/xrayprotocol.cpp
```diff
@@ -53,6 +54,18 @@ ErrorCode XrayProtocol::start()
 {
     qDebug() << "XrayProtocol::start()";
 
+    // Inject SOCKS5 auth into the inbound before starting xray.
+    // Re-uses existing credentials if the config already has them (e.g. imported config).
+    const auto credsOpt = amnezia::serialization::inbounds::EnsureInboundAuth(m_xrayConfig);
+    if (!credsOpt.has_value()) {
+        qCritical() << "Failed to acquire a free TCP port for local SOCKS inbound";
```
ygurov: This error is not clear from the method description and calling procedure
I would recommend just to call `throw` in case of any socket establishment issues with the specific error code/text in place there, then handle this issue in here with `try-catch`

client/core/serialization/inbound.cpp
```diff
@@ -14,25 +19,125 @@ namespace amnezia::serialization::inbounds
...
+            return i;
+    }
+    return -1;
+}
+
+// Ask the OS for a free TCP port on loopback (same stack as inbound "listen": "127.0.0.1").
+static std::optional<int> acquireFreeLocalPort()
+{
+    constexpr int kAttempts = 5;
+    for (int attempt = 0; attempt < kAttempts; ++attempt) {
```
ygurov: There is an only reason to do that IMO: we are out of ports range, which is quite uncommon, and which would break our further logic anyways.
All other errors are related to permissions issues, OS-specific network issues and so on.

So I think this process could be omitted without a doubt on all platforms. Did you encounter any of those in process of testing?

ygurov (changes requested):

ygurov (approved):

PR #2461, opened 2026-04-10
fix: dns resolve and reconnection issue

NickVs2015:
add vpn connection async
fix dns
fix reconnection when NM down/up actions
fix killswitch

### DIFF
--- a/client/platforms/linux/daemon/dnsutilslinux.cpp
+++ b/client/platforms/linux/daemon/dnsutilslinux.cpp
@@ -1,6 +1,3 @@
-/* This Source Code Form is subject to the terms of the Mozilla Public
- * License, v. 2.0. If a copy of the MPL was not distributed with this
- * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 #include "dnsutilslinux.h"
 
@@ -26,7 +23,6 @@ DnsUtilsLinux::~DnsUtilsLinux() {
   logger.debug() << "DnsUtilsLinux destroyed.";
 }
 
-
 void DnsUtilsLinux::writeResolvConf(const QList<QHostAddress>& resolvers) {
   if (resolvers.isEmpty()) return;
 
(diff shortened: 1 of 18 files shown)


client/platforms/linux/daemon/dnsutilslinux.cpp
```diff
@@ -4,219 +4,162 @@
...
-  m_ifindex = if_nametoindex(qPrintable(ifname));
-  if (m_ifindex <= 0) {
-    logger.error() << "Unable to resolve ifindex for" << ifname;
-    return false;
+void DnsUtilsLinux::writeResolvConf(const QList<QHostAddress>& resolvers) {
+  if (resolvers.isEmpty()) return;
+
+  static const QString kPath = QStringLiteral("/etc/resolv.conf");
+
+  if (m_resolvConfOriginal.isEmpty()) {
```
ygurov, line 34: Manual ops on resolve.conf is general a super bad idea. Modern linux manages that file automatically, and you even have some comments in the header states this "do not override" note.
This approach is wrong and must be avoided

client/vpnconnection.cpp
```diff
@@ -57,11 +57,15 @@ void VpnConnection::onKillSwitchModeChanged(bool enabled)
...
 #ifdef AMNEZIA_DESKTOP
     IpcClient::withInterface([enabled](QSharedPointer<IpcInterfaceReplica> iface){
-        QRemoteObjectPendingReply<bool> reply = iface->refreshKillSwitch(enabled);
-        if (reply.waitForFinished() && reply.returnValue())
-            qDebug() << "VpnConnection::onKillSwitchModeChanged: Killswitch refreshed";
-        else
-            qWarning() << "VpnConnection::onKillSwitchModeChanged: Failed to execute remote refreshKillSwitch call";
+        auto reply = iface->refreshKillSwitch(enabled);
+        auto *watcher = new QRemoteObjectPendingCallWatcher(reply);
+        QObject::connect(watcher, &QRemoteObjectPendingCallWatcher::finished, [watcher]() {
```
ygurov, line 62: Is there any point making this explicitly async? The same for the other similar calls

ygurov (changes requested):

PR #2479, opened 2026-04-14
fix: outbound freedom for xray 

cd-amn:


deploy/data/macos/pf/amn.400.allowPIA.conf
```diff
@@ -1,2 +1,2 @@
 # Allow traffic by privileged group (used by daemon)
-pass out proto { tcp, udp } group  { amnvpn } flags any no state
+pass out proto { tcp, udp } user  { root } flags any no state
```
ygurov: Using root user as a PF trigger generally a bad idea, what if something else would send some traffic under root?
Let's debug whether the amnvpn group is getting set for our app, and add some code in installers/build scripts

ygurov (approved):

Issue #2248, opened 2026-02-19
AmneziaVPN Windows very long connection establishing with long split tunneling config


hellchickens:
**Describe the bug**
Win client 4.8.12.9, WG2 protocol,  connects well and fast without split tunnel config. With imported split tunneling config with 11216 IP CIDRs records (full Russian IP CIDRs list) it takes app up to 5 minutes to  establish actual  VPN session with split tunneling working as expected, though application stays in "Connecting" state around 20 minutes until it gets in "Connected" state.

**To Reproduce**
1. Go to 'Site-based split tunneling'
2. Change mode to 'Addresses from the list should not be accessed via VPN'
3. Import very long list of CIDRs (example attached)
4. Enable 'Site-based split tunneling'
5. Connect to server

**Expected behavior**
Connection established in few second like on other platforms with same config. Tested on Mac, Android and iOS and it connects in 1-3 seconds.

**Log files**
```
[2026-02-19 19:57:57.775Z] [DEBUG] Amnezia  : Logger::setServiceLogsEnabled(): Logs transitioned to be  enabled
[2026-02-19 19:57:57.775Z] [INFO] Amnezia  : Logging has enabled on AmneziaVPN version 4.8.12.9 f7379284
[2026-02-19 19:57:57.775Z] [INFO] Amnezia  : Windows 11 Version 25H2 (x86_64)
[2026-02-19 19:58:22.496Z] [DEBUG] Amnezia  : VpnConfigurator::getDnsForConfig "1.1.1.1" "1.0.0.1"
[2026-02-19 19:58:22.496Z] [DEBUG] Amnezia  : "ConnectToVpn, Server index is 0, container is amnezia-awg2, route mode is" Settings::VpnAllExceptSites
[2026-02-19 19:58:22.499Z] [DEBUG] Amnezia LocalSocketController : Initializing
[2026-02-19 19:58:22.499Z] [DEBUG] Amnezia LocalSocketController : Connecting to: \\.\pipe\amneziavpn
[2026-02-19 19:58:22.499Z] [DEBUG] Amnezia LocalSocketController : Daemon connected
[2026-02-19 19:58:22.499Z] [DEBUG] Amnezia LocalSocketController : Check status
[2026-02-19 19:58:22.499Z] [DEBUG] Amnezia LocalSocketController : Reading
[2026-02-19 19:58:22.499Z] [DEBUG] Amnezia LocalSocketController : Parse command: status
[...]

Issue #2109, opened 2025-12-31
[UI/UX] `split tunneling` domain / ip groupping


VanillaMaster:
Add ability to:
- group ip / domains in [details](https://developer.mozilla.org/ru/docs/Web/HTML/Reference/Elements/details)-like groups, that can be expanded on demand (site split tunneling)
- toggle to enable / disable certain record or entire group (site split tunneling)

its very inconviniant to manage site split tunneling in case where its partialy provided by resources like [iplist](https://iplist.opencck.org/) and still managed by hand, due to total amount of records in list

Issue #1750, opened 2025-07-26
Site-based split tunneling uses only one IP per domain


qwelias:
**Describe the bug**
When specifying the domain name Site-based split tunneling uses only first IP addres among multiple A/AAAA DNS records

**To Reproduce**
1. Add `video-edge-1ac82c.pdx01.abs.hls.ttvnw.net` to Site-based split tunneling
2. Press ..., click "Save site list"
3. Open saved file, see one IP
4. Exec CLI `dig video-edge-1ac82c.pdx01.abs.hls.ttvnw.net` see multipe IP addresses

**Expected behavior**
All DNS records for the domain are added to the list

**Desktop (please complete the following information):**
 - OS: Linux
 - Version 4.8.7.2

Issue #1623, opened 2025-05-26
[FEATURE]  Split tunneling


igolikov:
Dear Amnezia team, is it possible to add feature that allows to exclude some networks from tunneling? 
Like everything except 192.168.50.0/24 will be tunneling via Amnezia. Right now when I start Amnezia VPN everything is routed via Amnezia, even other VPN connections

back-lacking:
this is offered within the client. 

![Image](https://github.com/user-attachments/assets/253bbdcd-381d-4adc-b481-f5c370cb06a9)

igolikov:
> this is offered within the client.

No, that is opposite:

1. Site-based tunneling routes traffic for specific domains(actually IPs) through Amnezia. Since websites often use multiple domains (e.g., chatgpt.com, openai.com), you must configure each domain individually.
2. This method performs poorly with sites protected by services like Cloudflare. Amnezia remembers the current IP address , but the IP can change unexpectedly within minutes

what I need is just opposite - route everything via VPN except some subnets 

![Image](https://github.com/user-attachments/assets/0e30695d-6108-4312-b880-1334da221d78)

Noxius87:
From what is currently implemented in AmneziaVPN - this mod looks like the most logical solution for your request. All traffic, except for the excluded sites, will go through VPN

![Image](https://github.com/user-attachments/assets/e90c12ed-7496-4bd0-ab68-4296f86642b0)

igolikov:
> this is offered within the client.

thanks for pointing me, have not noticed that yet


I have added 10.62.0.0/23 to that list (that seems to be excessive - that is private network and Amnezia should not affect such networks) - and it doesn't work, i noticed that Amnezia adds some routes when I run it

before
```
10.62.0.0/23 via 192.168.184.1 dev tun0 proto static metric 10 
```

after starting Amnezia
```
0.0.0.0/1 dev amn0 metric 1
128.0.0.0/1 dev amn0 metric 1
...
10.62.0.0/23 via 192.168.88.1 dev enp6s0 
10.62.0.0/23 via 192.168.184.1 dev tun0 proto static metric 10 
```


so Amnezia routes 10.62.0.0/23 via my physical nic with higher priority

burner1024:
It seems to work for me (100.64. range - tailscale).
However, if killswitch is enabled while using split tunneling with subnet exception, systemd-resolved spikes to 100% CPU and stays there. With killswitch disabled, CPU usage is normal.

[...]

PR #2506, opened 2026-04-21
Add server-managed split tunneling rules

nik65:
## Summary

This PR adds server-managed routing rules for site-based split tunneling.

A server admin can now configure a list of domains, IP addresses, and CIDR subnets that should **not** be accessed through the VPN for users of a specific server. The setting is available under:

`Settings -> Servers -> Server -> Services -> Server routing rules`

The placement is intentional: these rules are server policy, not just a local device preference. The rules are published for the VPN containers installed on that server, synced by clients that connect to that server, and shown in the existing Site-based split tunneling UI as locked managed entries.

The feature is scoped to the existing bypass mode: **"Addresses from the list should not be accessed via VPN"**. It does not add a server-managed allowlist/"only these through VPN" mode.

## Motivation

There are repeated requests around sharing, centralizing, and scaling split-tunneling rules.

The closest request is Discussion #1960, which describes this almost directly: a server admin defines one or more split-tunneling site lists and applies them to server users, while clients may still keep their own local blacklists/whitelists with priority rules.

Related requests and reports point to the same missing layer:

- #2449 asks for sharing split-tunneling settings between users.
- #1039 asks for easier import and automatic updating of large site lists instead of manual JSON-only management.
- #1423 is an umbrella issue for improving site-based split tunneling and consolidating duplicates.
- #2109 and Discussion #2418 describe the need to organize large domain/IP lists instead of forcing users to manage everything manually in one flat local list.
- #1607, Discussion #1939, and #1750 show that domain/subdomain/IP resolution behavior is a recurring pain point.
- #1623 asks for excluding networks/subnets from tunneling.
- #2248 shows the practical impact of very large CIDR lists on connection behavior.
- Discussion #2033 asks for whitelist-style routing behavior and clearer routing control.

This PR does not attempt to solve every split-tunneling runtime issue on every platform. It adds the missing configuration and distribution layer: a server-managed ruleset that can be maintained by the server admin and applied consistently by clients using that server.

## User-visible behavior

Server admins get a new **Server routing rules** service entry next to other server-level services.

From that page an admin can:

- add domains, IP addresses, and CIDR subnets to the server-managed bypass list;
- import and export the list using the same JSON shape as regular site split tunneling;
- publish the rules for clients of that server;
- optionally force site-based split tunneling for clients that have it disabled.

Client behavior:

- managed rules apply only while using the server where they are configured;
- managed entries are displayed in the normal Site-based split tunneling list;
- managed entries are locked/non-editable and visually separated from local user entries;
- local user entries remain editable and are not overwritten;
- clients using the opposite mode, "Only the sites listed here will be accessed through the VPN", keep their current behavior and are not switched automatically.

Force toggle behavior:

- if site split tunneling is disabled locally, the client enables bypass mode and applies the server-managed list;
- if the client already uses bypass mode, the server-managed list is merged with the local list;
- if the client already uses the opposite allowlist mode, the server-managed bypass list is not forced into that mode.

## Implementation details

- Adds server-managed routing rule storage to the settings/config layer.
- Publishes a `rules.json` payload for installed VPN containers, instead of publishing only for one default container.
- Adds client sync on connect, retry on failed/invalid sync, and a daily refresh interval for successful syncs.
- Keeps the rules sync endpoint reachable when split tunneling is active, so clients can refresh rules through the connected VPN path.
- Adds strict payload validation and limits for downloaded rules before persisting them locally.
- Supports domains, IP addresses, and CIDR subnets in managed rules.
- Preserves stored IP/CIDR data on import when it is already present in the exported JSON, and resolves missing domain data when rules are saved/imported.
- Handles server config updates so stale local managed rules are not kept forever after the server-side list is changed or cleared.
- Adds the QML page, services model entry, navigation, managed-sites model, and translations for the new UI.

## Compatibility

Older clients ignore the new managed-rules data and continue to use their existing local split-tunneling settings. They will still connect normally; they simply will not fetch, display, or apply server-managed routing rules until upgraded.

Existing local split-tunneling lists are preserved. Managed rules are merged into the effective routing list for the active server and are not copied into unrelated server profiles.

Plain WireGuard/export-style flows are kept separate from the server-managed sync path, so ordinary config export should continue to work without requiring these new fields.

## Scope / non-goals

This PR intentionally keeps the server-managed UI to bypass rules only: addresses from the managed list should not be accessed via VPN.

It does not add:

- a server-managed "only these addresses through VPN" mode;
- per-user server-side rule assignment;
- wildcard/regex rule syntax;
- a full subscription/update system for external third-party lists;
- a complete rewrite of platform-specific split-tunneling implementations.

Those can be considered separately. This PR focuses on the shared admin-managed ruleset and the client sync/apply pipeline.

## Validation

Local validation performed on Windows:

- `cmake --build deploy\build_64 --config Release --target AmneziaVPN --parallel`
- `git diff --check`

Android validation:

- C++ sources compiled for `arm64-v8a`.
- Gradle debug build completed successfully.
- The CMake `apk` target still exits at the known `androiddeployqt` expected-APK-path step after Gradle reports `BUILD SUCCESSFUL`.

## Commit breakdown

- `42bb69cc` feat: add server-managed routing rules sync
- `126a6f37` feat(ui): add server routing rules settings
- `acf4c26d` chore(i18n): update server routing rules translations


mishpro:
Looking forward for this on release. Top changes!

PR #2517, opened 2026-04-23
feat(cli): introduce headless amnezia-cli for desktop automation

vitkuz573:
## Why this PR exists
Amnezia had desktop flows and partial command handling, but not a true headless client suitable for terminal-first usage and automation.  
This PR adds a dedicated CLI binary that works with the same core models/controllers as the app, so behavior stays consistent.

## What this PR adds
- A new executable: `amnezia-cli` (`client/cli`)
- Full command surface for day-to-day operations:
  - VPN state and control: `status`, `connect`, `disconnect`
  - Daemon lifecycle: `daemon start|stop|status`
  - Server management: `servers list|show|add|import|remove|set-default|scan`
  - Country selection for API-backed configs: `countries list|set`
  - Container management: `containers list|set-default|remove`
  - Provisioning flows: `install server`, `install container`
  - Maintenance: `logs cleanup`
- CLI documentation: `client/cli/README.md`

## Implementation highlights
- Added a separate CMake target for CLI (`client/cli/CMakeLists.txt`) and wired it into the client build graph.
- Reworked several controller access paths to be safe in headless context:
  - network manager access no longer depends on UI-only app helpers
  - clipboard operations are guarded for GUI availability
- Result: shared business logic can be used both by desktop UI and by CLI without duplicating behavior.

## Verification
- `cmake --build build --target amnezia-cli -j4`
- `./build/client/cli/amnezia-cli --help`
- `./build/client/cli/amnezia-cli status --json`
- `./build/client/cli/amnezia-cli servers list --json`
- `./build/client/cli/amnezia-cli countries list --index 0 --json`

### DIFF
--- a/client/CMakeLists.txt
+++ b/client/CMakeLists.txt
@@ -238,3 +238,5 @@ if(COMMAND qt_finalize_executable)
 else()
     qt_finalize_target(${PROJECT})
 endif()
+
+add_subdirectory(cli)
--- a/client/cli/CMakeLists.txt
+++ b/client/cli/CMakeLists.txt
@@ -1,0 +1,72 @@
+set(CLI_PROJECT amnezia-cli)
+
+set(CLI_SOURCES ${SOURCES})
+list(FILTER CLI_SOURCES EXCLUDE REGEX "/main\\.cpp$")
+list(FILTER CLI_SOURCES EXCLUDE REGEX "/amnezia_application\\.cpp$")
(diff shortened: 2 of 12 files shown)


WkdXeqtr:
I've been looking for something like this for a while. Having a CLI to connect/disconnect, switch servers, and control everything programmatically makes Amnezia actually usable in scripts and automation setups. Really hoping this gets some attention from maintainers - any chance of a review?

PR #2555, opened 2026-05-01
Support locations change

cd-amn:


client/core/tunnel.h
```diff
@@ -0,0 +1,82 @@
...
+#include <QSharedPointer>
+#include <QString>
+
+#include "core/protocols/vpnProtocol.h"
+#include "core/utils/containerEnum.h"
+#include "core/utils/errorCodes.h"
+
+class QTimer;
+
+class Tunnel : public QObject {
```
ygurov, line 15: Let's consider making this class to be inherited by VpnProtocol, or putting the whole code into VpnProtocol itself. I doubt we will use it separately from VpnProtocol itself as well as using VpnProtocol itself without these features.

client/core/vpnTrafficGuard.cpp
```diff
@@ -0,0 +1,452 @@
...
+            } else {
+                qDebug() << "VpnTrafficGuard::applyKillSwitch: Successfully enabled killswitch";
+            }
+        }
+#endif
+        const QString proto = updatedConfig.value(configKey::vpnProto).toString();
+        const bool isXrayBased = (proto == ProtocolUtils::protoToString(Proto::Xray) ||
+                                  proto == ProtocolUtils::protoToString(Proto::SSXray));
+        if (isXrayBased) {
+            if (updatedConfig.value(configKey::splitTunnelType).toInt() == amnezia::route_mode_ns::VpnAllSites) {
```
ygurov: Let's handle it in the setupRoutes call.

The lines below seems to be a legacy ones, since we have bind and/or routing holes for tunnel connection in all protocols. I think we can handle this route in the unified way for all the protocols and endpoints

client/core/vpnTrafficGuard.cpp
```diff
@@ -0,0 +1,452 @@
...
+        }
+    });
+#endif
+}
+
+void VpnTrafficGuard::flushAll()
+{
+#ifdef AMNEZIA_DESKTOP
+    IpcClient::withInterface([&](QSharedPointer<IpcInterfaceReplica> iface) {
+        iface->restoreTunnelResolvers();
```
ygurov: All replica calls are actually async. Checking the output code and waiting for the call as the one below is crucial. Otherwise, callchain turns out to be very unreliable and UB-risky

client/core/vpnTrafficGuard.cpp
```diff
@@ -0,0 +1,452 @@
...
+        QRemoteObjectPendingReply<bool> reply = iface->disableKillSwitch();
+        m_allowedEndpoints.clear();
+        //TODO: why it takes so long?
+        if (!reply.waitForFinished(1000) || !reply.returnValue()) {
```
ygurov: Let's remove all the specific timings from waitForFinished calls. If it takes that long - something is not right and we should address it somehow

PR #2571, opened 2026-05-05
feat: Add Korean language support

Nergis0318:
Hello, I’m a user from South Korea. Due to the recent increase in internet censorship in South Korea, I’ve started using Amnezia. I’m satisfied with everything, but I was disappointed that there was no Korean translation, so I’ve added one. I would appreciate it if you could accept this PR.


vkamn:
added here https://github.com/amnezia-vpn/amnezia-client/pull/2999

PR #2588, opened 2026-05-11
feat: add MasterDnsVPN protocol (native C++ engine)

coffeegrind123:
Adds MasterDnsVPN as a third SOCKS5-proxy-mode transport, alongside the existing Xray and SSXray entries. The four L3 transports (OpenVPN, WireGuard, AmneziaWG, IKEv2) are unaffected. MasterDnsVPN tunnels traffic inside DNS queries against operator-controlled tunnel domains. It's useful where standard VPN ports are blocked but DNS still resolves.

The reference implementation is masterking32/MasterDnsVPN, written in Go. Instead of bundling the Go binary, this PR ports the client engine to native C++/Qt6. The port integrates with the existing service/IPC architecture, ships statically with the rest of the client, and works on every platform amnezia-client already supports, without adding a per-platform Go toolchain to the build matrix.

## What landed

74 files, ~14.5k LOC, across three layers.

### UI and config model

QML page `PageProtocolMasterDnsVpnSettings.qml` for tunnel domains, resolvers, encryption method, SOCKS5 listen port and credentials. Backed by `MasterDnsVpnConfigModel` (structured JSON, no QSettings legacy strings). Plugs into the existing protocol dropdown via `protocolsModel`. `installUiController` is extended for SSH-installed servers.

### Native engine

Under `client/masterdnsvpn/`, 13 modules totaling ~8.3k LOC:

- `wireframing`: inner-packet binary codec, packed-control-blocks, SESSION_ACCEPT payload with the 13-byte client-policy tail.
- `crypto`: None, XOR, ChaCha20, AES-128/192/256-GCM cipher pair.
- `dnsframing`: base32/base36 codecs, RFC 1035 query builder, chunked TXT-RR response parser, exported DNS qtype/rcode/class constants.
- `compression`: ZSTD, LZ4, ZLIB-raw with a 10 MiB decompression-bomb cap.
- `arq`: per-stream reliability. Sliding-window, bounded NACK with frontier sampling, Karn-eligible adaptive RTO on both data and control planes, control-plane RTO retransmits, initial-NACK-delay, receive-window cap at 2x send window.
- `resolverpool`: one QUdpSocket per resolver, 8 balancing strategies (RoundRobin through LossThenLatency to LeastLossTopRandom/RoundRobin), `reportSend`/`reportSuccess`/`reportTimeout` feedback API, per-resolver MTU and rolling loss/RTT EWMA with half-life decay.
- `mtuprober`: binary-search upload and download MTU discovery.
- `pingpacer`: tiered ping-pacing FSM (aggressive/lazy/cooldown/cold).
- `socks5server`: local SOCKS5 listener for CONNECT (TCP) and UDP_ASSOCIATE (DNS-only, port 53). Standalone `parseTargetPayload` and UDP datagram codecs.
- `dnsmsg`: lite RFC 1035 parser. Extracts transaction-ID and the first question for cache keying. Tolerates pointer compression with a loop guard.
- `dnscache`: local DNS cache (Miss/Pending/Ready tri-state, configurable TTL) plus a DNS_QUERY_RES fragment reassembly store.
- `session`: dispatcher, SESSION_INIT/ACCEPT handshake, `applyServerPolicy` clamping, per-stream ArqStream lifecycle, DNS_QUERY_REQ/RES dispatch and cache integration.
- `engine`: façade for the desktop service host and Android JNI bridge.

### Service and IPC integration

Service-side `master_dns_vpn_service.{cpp,h}` host. IPC interface extended via `ipc_interface.rep` with `masterDnsVpnStart`, `masterDnsVpnStop`, and `masterDnsVpnSocksPort` slots. `conanfile.py` adds `zstd/1.5.6` and `lz4/1.10.0` for the compression codecs (zlib is already a project dep). No new platform-specific toolchain deps.

## Spec and tests

The wire spec is `docs/masterdnsvpn-wire-spec.md` (~800 lines). It walks the on-the-wire behavior section by section against upstream's Go source, with §13 citing the relevant `internal/arq`, `internal/vpnproto`, and `internal/client` files. §11 covers the local DNS service; §11.1a documents the deliberate divergence from upstream described below.

Tests are in `client/tests/testMasterDnsVpnEngine.cpp` (~3.3k LOC). Most entries are direct translations of upstream's Go test suite (`internal/arq/arq_test.go`, `internal/vpnproto/*_test.go`, `internal/basecodec/*_test.go`, `internal/security/codec_test.go`, `internal/client/balancer_test.go`, `internal/fragmentstore/store_test.go`, etc.) into QTest. Two real engine semantic bugs in the C++ port (HybridScore and LeastLossTop\* round-robin fallback) turned up during the parity sweep and were fixed in flight.

## One deliberate divergence: SOCKS5 cache-miss

Wire protocol matches upstream Go byte for byte. Server interop is preserved. One client-internal behavior diverges, documented in §11.1a of the spec:

Upstream's pattern on a SOCKS5-UDP DNS cache miss is: dispatch the query to the tunnel, then close the SOCKS5 UDP association so the local app retries, and let the retry hit the now-populated cache. This port instead dispatches, tracks the in-flight query by wire seq, and sends the tunnel response back when it arrives. No association teardown.

The reason: closing the listening socket on cache miss is unique to MasterDnsVPN. Normal DNS resolvers (dnsmasq, systemd-resolved, unbound) don't do that. The socket-churn pattern is observable on the local machine and is enough to fingerprint the resolver. The C++ port behaves like a standard cache-then-forward resolver and is indistinguishable from dnsmasq at the loopback layer. On-wire OPSEC is unchanged either way; local-loopback observability improves.

## Out of scope

Client-only. Operators run the existing Go server on the VPS. No DNS-response builders, no server-side fragment store, no UDP server-side bind paths.

Pool reconfiguration after `start()` isn't supported, matching the other protocol implementations in this client. Resetting means tearing down and starting a fresh session.

## Compatibility

No changes to existing protocols, models, or wire formats. MasterDnsVPN shows up as a new entry in the proxy-transport dropdown. Existing configurations are unaffected. The only new runtime dependencies are zstd and lz4 via Conan.

If splitting this into smaller PRs (UI + model, engine, service, spec, tests) would be easier to review, that's fine.


coffeegrind123:
Closing — we're keeping MasterDnsVPN in our own fork for now rather than upstreaming it.

PR #2601, opened 2026-05-14
Build macOS prebuilts for arm64 and x86_64

yyy-amnezia:


recipes/awg-go/conanfile.py
```diff
@@ -21,12 +24,84 @@ def _goos(self):
...
+        return os.path.join(self.dependencies.build["go"].package_folder, "bin")
+
+    @property
+    def _macos_deployment_target(self):
+        if str(self.settings.os) != "Macos":
+            return None
+        return self.settings.get_safe("os.version")
+
+    def _go_env(self, goarch):
+        env = Environment()
```
ygurov: Do not use Environment manually, or not as a main one - it ignores externally-specified flags and Conan internals passing to the Makefile.

Use an env on a top of this one instead
```
tc = AutotoolsToolchain(self)
env = tc.environment()
```

recipes/awg-go/conanfile.py
```diff
@@ -21,12 +24,84 @@ def _goos(self):
...
+        env.define("GOOS", self._goos)
+        env.define("GOARCH", goarch)
+        env.define("GOROOT", self.dependencies.build["go"].package_folder)
```
ygurov: Do not specify GOROOT - it does not change the actual go executable used, just replaces the toolchain location.
I see we have issues with passing our external go in here, so let's use this one as a first line in `def generate(self)`. Should resolve the issue
```
VirtualBuildEnv(self).generate()
```

recipes/awg-go/conanfile.py
```diff
@@ -21,12 +24,84 @@ def _goos(self):
...
+        env.define("GOPATH", os.path.join(self.build_folder, "gopath"))
+        env.define("GOMODCACHE", os.path.join(self.build_folder, "gopath", "pkg", "mod"))
```
ygurov: Hmm, specifying this one is not a bad idea!
Let's do it for all platforms, should make builds more consistent

recipes/awg-go/conanfile.py
```diff
@@ -21,12 +24,84 @@ def _goos(self):
...
+        env.define("GOCACHE", os.path.join(self.build_folder, "gocache"))
+        env.define("GOTELEMETRY", "off")
+        if self._macos_deployment_target:
+            env.define("MACOSX_DEPLOYMENT_TARGET", self._macos_deployment_target)
+        env.prepend_path("PATH", self._go_bin_path)
+        return env
+
+    def _cgo_arch_flags(self, goarch):
+        return f"-arch {self._clang_arch_map[goarch]}"
```
ygurov: `-arch` option used in here allows values in apple-clang-internal format, not the golang one.
These values are corresponding with the ones used in Conan, so let's just use them

Moreover, Conan does specify multiple `-arch` flags itself during `generate` step if `AutotoolsToolchain` is used.

recipes/awg-go/conanfile.py
```diff
@@ -21,12 +24,84 @@ def _goos(self):
...
+        env = Environment()
+        env.define("GOOS", self._goos)
+        env.define("GOARCH", goarch)
+        env.define("GOROOT", self.dependencies.build["go"].package_folder)
+        env.define("GOPATH", os.path.join(self.build_folder, "gopath"))
+        env.define("GOMODCACHE", os.path.join(self.build_folder, "gopath", "pkg", "mod"))
+        env.define("GOCACHE", os.path.join(self.build_folder, "gocache"))
+        env.define("GOTELEMETRY", "off")
+        if self._macos_deployment_target:
+            env.define("MACOSX_DEPLOYMENT_TARGET", self._macos_deployment_target)
```
ygurov: This one would be specified automatically if `AutotoolsToolchain` is used. I would suggest just make a proper generate step instead of applying every flag manually

recipes/awg-go/conanfile.py
```diff
@@ -21,12 +24,84 @@ def _goos(self):
...
+        env.prepend_path("PATH", self._go_bin_path)
```
ygurov: That's a bad idea. Use the one suggested in
https://github.com/amnezia-vpn/amnezia-client/pull/2601/changes#r3247262682

recipes/awg-go/conanfile.py
```diff
@@ -21,12 +24,84 @@ def _goos(self):
...
+        return env
+
+    def _cgo_arch_flags(self, goarch):
+        return f"-arch {self._clang_arch_map[goarch]}"
+
+    def _build_go_arch(self, goarch):
```
ygurov: DRY
Let's use an existing build step iterating all over archs, and use lipo then

recipes/awg-go/conanfile.py
```diff
@@ -21,12 +24,84 @@ def _goos(self):
...
+        env = self._go_env(goarch)
+        env.define("CGO_CFLAGS", self._cgo_arch_flags(goarch))
+        env.define("CGO_LDFLAGS", self._cgo_arch_flags(goarch))
+        with env.vars(self).apply():
+            self.run("make", env=None)
```
ygurov: Do not run make manually, use
```
at = Autotools(self)
at.make()
```

Otherwise, it would ignore all Conan internal-specified variables

ygurov (changes requested):
Left some comments for a single recipe though to be applied for all modified recipes

client/CMakeLists.txt
```diff
@@ -252,6 +252,25 @@ if(OVPN_SCRIPTS)
         "$<TARGET_FILE_DIR:${PROJECT}>"
     )
 
+    if(APPLE AND NOT IOS AND NOT MACOS_NE)
+        set(SIGN_OVPN_SCRIPT [=[
+if [ "$CODE_SIGNING_ALLOWED" != "NO" ]; then
```
ygurov: This is a huge hack that should be avoided at any cost - very fragile and non-transparent.
You can review the process of signing the binary from CPack cmake code, there should be more appropriate place for the code

recipes/awg-go/conanfile.py
ygurov: I still have a complaint about the building process - there should be a single function which does the job.
I find trying to put something into the hooks a very bad practice. Recipe becomes unclear to read and analyze in the future
DRY KISS

recipes/awg-go/conanfile.py
```diff
@@ -21,12 +26,70 @@ def _goos(self):
...
+            "GOCACHE": os.path.join(self.build_folder, "gocache"),
+            "GOTELEMETRY": "off",
+        }
+
+    def _define_go_cache_env(self, env):
+        for name, value in self._go_cache_vars().items():
+            env.define(name, value)
+
+    def _go_arch_make_args(self, goarch):
+        return [f"{name}={value}" for name, value in self._go_cache_vars().items()] + [
```
ygurov: Looks like a slope. Why don't we just add them by env.define() calls?
Complex logic is not what we want here
KISS

recipes/awg-go/conanfile.py
```diff
@@ -21,12 +26,70 @@ def _goos(self):
...
+    def _go_cache_vars(self):
+        return {
+            "GOPATH": os.path.join(self.build_folder, "gopath"),
+            "GOMODCACHE": os.path.join(self.build_folder, "gopath", "pkg", "mod"),
+            "GOCACHE": os.path.join(self.build_folder, "gocache"),
+            "GOTELEMETRY": "off",
+        }
+
+    def _define_go_cache_env(self, env):
+        for name, value in self._go_cache_vars().items():
```
ygurov: Same for here. Why do we need all of those abstractions if we can define those vars in place at generate() stage?
KISS

recipes/awg-go/conanfile.py
```diff
@@ -46,22 +109,29 @@ def source(self):
         )
 
     def generate(self):
+        VirtualBuildEnv(self).generate()
```
ygurov: All recipes have virtualbuildenv by default, if it's not turned off explicitly. Redundant; safe to remove

recipes/awg-go/conanfile.py
```diff
@@ -21,12 +26,70 @@ def _goos(self):
...
+
+    def _go_arch_make_args(self, goarch):
+        return [f"{name}={value}" for name, value in self._go_cache_vars().items()] + [
+            f"GOOS={self._goos}",
+            f"GOARCH={goarch}",
+        ]
+
+    def _build_go_arch(self, goarch):
+        autotools = Autotools(self)
+        autotools.make(args=self._go_arch_make_args(goarch))
```
ygurov: Subsequent calls of make with different arch could lead to UB. Make sure it's cleared or build in separate folders

recipes/tun2socks/conanfile.py
```diff
@@ -22,17 +26,82 @@ def _goos(self):
...
+    @property
+    def _goarchs(self):
+        return [self._arch_map.get(arch) for arch in self._archs]
+
+    @property
+    def _is_universal_macos(self):
+        return str(self.settings.os) == "Macos" and len(self._archs) > 1
+
+    @property
+    def _is_unsupported_multi_arch(self):
```
ygurov: Property is overkill, could be inlined. KISS

ygurov (changes requested):
Once again, reviewed just a single recipe, but to be applied for every changed one

client/CMakeLists.txt
```diff
@@ -239,18 +239,33 @@ install(SCRIPT ${QT_DEPLOY_SCRIPT}
...
+        MACOSX_PACKAGE_LOCATION MacOS
+    )
+    target_sources(${PROJECT} PRIVATE ${MACOS_OVPN_SCRIPT})
+    if(BUILD_VPN_KEYCHAIN)
+        set(MACOS_CODE_SIGN_FLAGS "--keychain ${BUILD_VPN_KEYCHAIN} --deep")
+    else()
+        set(MACOS_CODE_SIGN_FLAGS "--deep")
+    endif()
+    set_target_properties(${PROJECT} PROPERTIES
+        XCODE_ATTRIBUTE_OTHER_CODE_SIGN_FLAGS "${MACOS_CODE_SIGN_FLAGS}"
```
ygurov: Macos build could be assembled using not only the Xcode generator. I find this one an unreliable thing

client/CMakeLists.txt
```diff
@@ -239,18 +239,33 @@ install(SCRIPT ${QT_DEPLOY_SCRIPT}
...
 
 if (APPLE AND NOT IOS AND NOT MACOS_NE)
-    list(APPEND OVPN_SCRIPTS "${CMAKE_SOURCE_DIR}/deploy/data/macos/update-resolv-conf.sh")
+    set(MACOS_OVPN_SCRIPT "${CMAKE_SOURCE_DIR}/deploy/data/macos/update-resolv-conf.sh")
+    list(APPEND OVPN_SCRIPTS ${MACOS_OVPN_SCRIPT})
+    set_source_files_properties(${MACOS_OVPN_SCRIPT} PROPERTIES
+        MACOSX_PACKAGE_LOCATION MacOS
+    )
+    target_sources(${PROJECT} PRIVATE ${MACOS_OVPN_SCRIPT})
+    if(BUILD_VPN_KEYCHAIN)
```
ygurov: Why do we sign something during build? Please consider doing that in CPack as I mentioned previously. Or let's discuss that if you do not agree

recipes/awg-go/conanfile.py
ygurov: This one is still unresolved

recipes/awg-go/conanfile.py
ygurov: This one as well

ygurov (changes requested):

client/CMakeLists.txt
```diff
@@ -239,18 +239,29 @@ install(SCRIPT ${QT_DEPLOY_SCRIPT}
...
     list(APPEND OVPN_SCRIPTS "${CMAKE_SOURCE_DIR}/deploy/data/linux/update-resolv-conf.sh")
 endif()
 
 if(OVPN_SCRIPTS)
-    add_custom_command(TARGET ${PROJECT} POST_BUILD
-        COMMAND ${CMAKE_COMMAND} -E copy_if_different
-        ${OVPN_SCRIPTS}
-        "$<TARGET_FILE_DIR:${PROJECT}>"
-    )
+    if(NOT APPLE)
```
ygurov: I assume on Apple it should be copied as well

client/CMakeLists.txt
```diff
@@ -239,18 +239,29 @@ install(SCRIPT ${QT_DEPLOY_SCRIPT}
 )
 
 if (APPLE AND NOT IOS AND NOT MACOS_NE)
-    list(APPEND OVPN_SCRIPTS "${CMAKE_SOURCE_DIR}/deploy/data/macos/update-resolv-conf.sh")
+    set(MACOS_OVPN_SCRIPT "${CMAKE_SOURCE_DIR}/deploy/data/macos/update-resolv-conf.sh")
```
ygurov: Do we really need this change? :D
Looks like we use an additional variable to store a single string in the list

recipes/awg-go/conanfile.py
```diff
@@ -48,20 +69,63 @@ def source(self):
...
         env.define("CGO_CFLAGS", tc.cflags)
         tc.generate(env)
 
     def build(self):
+        if self._is_universal_macos:
+            outputs = []
+            for goarch in self._goarchs:
+                arch_build_folder = os.path.join(self.build_folder, f"build-{goarch}")
+                rmdir(self, arch_build_folder)
+                copy(self, "*", src=self.source_folder, dst=arch_build_folder, excludes=(
```
ygurov: Looks complicated and untrivial.
I would suggest to use DESTDIR variable of the original Makefile

recipes/awg-go/conanfile.py
```diff
@@ -48,20 +69,63 @@ def source(self):
...
+                    "conan",
+                    "conan/*",
+                    "gocache",
+                    "gocache/*",
+                    "gopath",
+                    "gopath/*",
+                ))
+
+                at = Autotools(self)
+                with chdir(self, arch_build_folder):
```
ygurov: This one ain't gonna work. To execute Makefile you have to be in the same directory with it, or use `-C` option. The 2nd could produce unexpected output

recipes/awg-go/conanfile.py
```diff
@@ -48,20 +69,63 @@ def source(self):
...
         tc = AutotoolsToolchain(self)
         env = tc.environment()
+        env.define("GOPATH", os.path.join(self.build_folder, "gopath"))
+        env.define("GOMODCACHE", os.path.join(self.build_folder, "gopath", "pkg", "mod"))
+        env.define("GOCACHE", os.path.join(self.build_folder, "gocache"))
+        env.define("GOTELEMETRY", "off")
         env.define("GOOS", self._goos)
-        env.define("GOARCH", self._goarch)
+        if not self._is_universal_macos:
+            env.define("GOARCH", self._goarch)
```
ygurov: You already define one in Makefile call itself. I do not think the extra one is required

recipes/awg-go/conanfile.py
```diff
@@ -48,20 +69,63 @@ def source(self):
...
+                    "gocache",
+                    "gocache/*",
+                    "gopath",
+                    "gopath/*",
+                ))
+
+                at = Autotools(self)
+                with chdir(self, arch_build_folder):
+                    at.make(args=[
+                        f"GOOS={self._goos}",
```
ygurov: I do not think multi-os bundles are possible :)
Let's stick with the only one defined in env

recipes/awg-go/conanfile.py
```diff
@@ -48,20 +69,63 @@ def source(self):
...
         env.define("GOOS", self._goos)
-        env.define("GOARCH", self._goarch)
+        if not self._is_universal_macos:
+            env.define("GOARCH", self._goarch)
         env.define("CGO_LDFLAGS", tc.ldflags)
         env.define("CGO_CFLAGS", tc.cflags)
         tc.generate(env)
 
     def build(self):
+        if self._is_universal_macos:
```
ygurov: Still special handling in here.
Let's use a unified way to compile the binary despite the fact it is multiarch. The only difference is the merging thingy made by lipo

ygurov (changes requested):

recipes/awg-go/conanfile.py
```diff
@@ -21,12 +25,29 @@ def _goos(self):
...
             "armv8": "arm64"
-        }.get(str(self.settings.arch))
+        }
+
+    @property
+    def _archs(self):
+        return str(self.settings.arch).split("|")
+
+    @property
+    def _goarch(self):
```
ygurov: Seems to be unused, let's delete it

recipes/awg-go/conanfile.py
```diff
@@ -21,12 +25,29 @@ def _goos(self):
...
+        goarchs = [self._arch_map.get(arch) for arch in self._archs]
+        return goarchs[0] if len(goarchs) == 1 else None
+
+    @property
+    def _goarchs(self):
+        return [self._arch_map.get(arch) for arch in self._archs]
+
+    @property
+    def _is_universal_macos(self):
```
ygurov: Please make it in the same way as the one in amnezia-xray-bindings, or change it there.
Consistency

recipes/tun2socks/conanfile.py
```diff
@@ -66,23 +78,46 @@ def source(self):
...
+
+        os.rename(outputs[0], output)
 
     def package(self):
-        copy(self, "tun2socks", src=self.build_folder, dst=self.package_folder)
+        copy(self, self._binary_name, src=self.build_folder, dst=self.package_folder)
         if self._is_windows:
             with chdir(self, self.package_folder):
-                os.rename(src="tun2socks", dst="tun2socks.exe")
+                os.rename(src=self._binary_name, dst=f"{self._binary_name}.exe")
```
ygurov: What's the point of renaming the binary twice?
The one in here, and the one in `os.rename(output_path, arch_output_path)`

recipes/tun2socks/conanfile.py
```diff
@@ -41,7 +53,7 @@ def layout(self):
         basic_layout(self)
 
     def validate(self):
-        if not self._goos or not self._goarch:
+        if not self._goos or not all(self._goarchs) or (len(self._archs) > 1 and not self._is_universal_macos):
```
ygurov: Please make separated explicit error about the multiarch support

recipes/awg-go/conanfile.py
```diff
@@ -48,20 +69,48 @@ def source(self):
     def generate(self):
         tc = AutotoolsToolchain(self)
```
ygurov, line 68: AutotoolsToolchain automatically applies arch-specific flags to cflags and ldflags. Please refer to `amnezia-xray-bindings`, the place with `apple_arch_flag`

recipes/awg-go/conanfile.py
```diff
@@ -48,20 +69,48 @@ def source(self):
...
 
     def build(self):
-        at = Autotools(self)
-        at.make()
+        outputs = []
+        with chdir(self, self.source_folder):
+            for goarch in self._goarchs:
+                arch_destdir = os.path.join(self.build_folder, f"build-{goarch}")
+                at = Autotools(self)
+                at.make("clean")
```
ygurov: `Make` does not receive any of cflags and ldflags. Providing them is crucial

recipes/awg-go/conanfile.py
```diff
@@ -48,20 +69,48 @@ def source(self):
...
+                at.make("install", args=[
+                    f"DESTDIR={shlex.quote(arch_destdir)}",
+                    "BINDIR=",
+                    f"GOARCH={goarch}",
+                ])
+                output_path = os.path.join(arch_destdir, self._binary_name)
+                arch_output_path = os.path.join(self.build_folder, f"{self._binary_name}-{goarch}")
+                os.rename(output_path, arch_output_path)
```
ygurov: What's the point of renaming of the binary if DESTDIR is specified? They all belong to separate dirs

recipes/awg-go/conanfile.py
```diff
@@ -48,20 +69,48 @@ def source(self):
...
+                " ".join(shlex.quote(output) for output in outputs),
+                shlex.quote(output)
+            ))
+            return
+
+        os.rename(outputs[0], output)
 
     def package(self):
-        copy(self, "amneziawg-go", src=self.build_folder, dst=self.package_folder)
+        copy(self, self._binary_name, src=self.build_folder, dst=self.package_folder)
```
ygurov: That's not going to work for any non-apple

ygurov (changes requested):

PR #2649, opened 2026-05-24
Ubuntu22-compatible build (Dockerfile)

Zaburunier:
A few latest releases of AmneziaVPN are incompatible with Ubuntu 22.04. I needed VPN for my home PC which runs exactly 22.04, so I made this Dockerfile to address the issue.

### DIFF
--- a/docker/ubuntu-22.04.Dockerfile
+++ b/docker/ubuntu-22.04.Dockerfile
@@ -1,0 +1,184 @@
+# syntax=docker/dockerfile:1
+# escape = \
+
+
+ARG QT_INSTALLATION_PREFIX="/opt/Qt"
+ARG QT_INSTALLATION_RELEASE="6.11.1"
+ARG QIF_INSTALLATION_RELEASE="4.11"
+ARG QIF_INSTALLATION_RELEASE_CODE="411"
+
+ARG CONAN_OUTPUT_FOLDER="/opt/conan"
+
+ARG PROJECT_WORKDIR="/opt/amnezia/client"
+
+# Switch from "remote" to "local" to reuse local directory instead of using "git clone"
+# NOTE: when using local directory, "docker build" must be called from repository root!
+ARG USE_REPO="remote"
+ARG REPO_URL="https://github.com/amnezia-vpn/amnezia-client"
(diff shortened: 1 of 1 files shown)


Zaburunier:
Things to note:
1. At the moment, Qt official installer is used. This means, valid Qt account is required. This is done by sharing qtaccount.ini file with container via `--mount=type=secret` in Dockerfile. Also login/password can be passed as build arguments but this is insecure and thus disabled (commented out) by default.
2. Since I don't have access to Amnezia's conan remote, I had to use the following workaround:
- Prepare conan packages outside CMake build process using default remote _conancenter_ (see stage 3 in Dockerfile)
- Pass `--no-remote` to CMake build process to force local cache usage (see toolchain file in stage 5)
3. This Dockerfile can be easily extended to any Linux distribution. The single distro-specific thing I used is `apt-get`. CMake and Ninja is taken from Qt installation. Some very old distro can be selected to provide widest possible compatibility range (glibc release etc.). The only question is: will it be useful for you?

Zaburunier:
Launch command example:
```bash
DOCKER_BUILDKIT=1 docker build --secret id=qt_credentials,src=$HOME/.local/share/Qt/qtaccount.ini --file docker/ubuntu-22.04.Dockerfile --tag amnezia-ubuntu22 .
```

Then, to retrieve installation package:
```bash
docker container create --name amnezia-ubuntu22-container amnezia-ubuntu22 .
docker container cp amnezia-ubuntu22-container:/dist/. ./.build-dockerfile/
docker container rm amnezia-ubuntu22-container
```

---

Do everything with a single command:
```bash
DOCKER_BUILDKIT=1 docker build --secret id=qt_credentials,src=$HOME/.local/share/Qt/qtaccount.ini --file docker/ubuntu-22.04.Dockerfile --tag amnezia-ubuntu22 --output=./.build-dockerfile/ .
```

Zaburunier:
I know that releases are currently built the other way, this is just draft from personal use case. I am totally open for discussion, maybe there is something I can be helpful with.

Zaburunier:
Написал небольшой пост по теме на русском: https://habr.com/ru/posts/1043898/

PR #2684, opened 2026-06-01
Fix Fedora Linux reconnect after suspend

pxc1984:
I had a very specific issue with AmneziaVPN client on my linux laptop: after sleep mode the amneziavpn client broke the network, spiked the cpu and required to 
manually reconenct to the server for the network to be gone.

dove deeper into the code and found out that amneziavpn daemon was trying to recreate itself way before network was up and it stalled into some firewall/route/cgroup 
fedora-specific incompatabilities and left the system in partially broken state

- replace the named routing table amnvpnrt with numeric table id 3211
- skip net_cls handling on cgroup v2 systems
- create the missing 400.allowPIA firewall anchor before using it
- fix 120.blockNets handling to enable the correct killswitch rule
- emit wakeup/reconnect only after NetworkManager reaches CONNECTED_GLOBAL

### DIFF
--- a/client/platforms/linux/daemon/linuxfirewall.cpp
+++ b/client/platforms/linux/daemon/linuxfirewall.cpp
@@ -33,6 +33,7 @@
 #include "linuxfirewall.h"
 #include "logger.h"
 #include "xray_defs.h"
+#include <QFileInfo>
 #include <QProcess>
 
 #define BRAND_CODE "amn"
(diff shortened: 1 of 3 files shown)


pxc1984:
this solved the issue for me, so i'll continue using my locally built binary, but just wanted to save the day for anyone else having a similar issue)

PR #2727, opened 2026-06-13
Fix: enable 120.blockNets by blockNets on Linux (split-tunnel)

yp-amnezia:
…nnel "only forward sites" leak).


vkamn:
merged here https://github.com/amnezia-vpn/amnezia-client/pull/2820

Issue #2586, opened 2026-05-08
Windows: app-based split tunneling doesn't work


yupic:
**Describe the bug**
On Windows, app-based split tunneling does not work correctly, e.g. excluded applications don't bypass VPN when accessing local network resources.

If Chrome is added to the app-based split tunneling exclusion list, blocked in Russia sites may still open, but local LAN resources such as a router web interface at `192.168.0.1` do not open.

The issue is protocol-independent: it happens with both VLESS/Xray and AmneziaWG.

**To Reproduce**
1. Disable site-based split tunneling.
2. Enable app-based split tunneling with `Apps from the list should not have access via VPN`.
3. Add `chrome.exe` to the exclusion list.
4. Connect to VPN.
5. Open some blocked site, for example youtube.com in Chrome.
6. YouTube opens.
7. Open a local LAN resource in Chrome, for example:
   ```text
   http://192.168.0.1
   ```
8. The router web interface does not open.

**Expected behavior**

An application excluded from VPN should bypass VPN completely.

Chrome should not be able to access to blocked site.

Chrome should be able to access to local LAN resources such as `192.168.0.1`.

Traffic from excluded apps should use the physical network interface, not the VPN/TUN interface.

**Log files**
None

**Screenshots**
A screenshot from Windows Resource Monitor shows that connections from `chrome.exe` are made through the VPN/TUN interface `10.33.0.2`, even though Chrome is added to app-based split tunneling exclusions.

<img width="2880" height="1740" alt="Image" src="https://github.com/user-attachments/assets/328cff9b-9d62-4378-88b8-f11e58ac3672" />

**Desktop (please complete the following information):**
- OS: Microsoft Windows 11 Home Single Language
- OS version: 10.0.26200.8246, Build 26200, 25H2
- Architecture: x64
- AmneziaVPN Version: 4.8.15.4
- AmneziaVPN Service Version: 4.8.15.4

**Smartphone (please complete the following information):**
Not applicable.

**Server (please complete the following information):**
Protocol-independent. Reproduced with:
- VLESS/Xray
- AmneziaWG

danielstuffs:
Happens on my end aswell

Foulvertex:
Такая же история. После введения блокировок из-за бугра на некоторых отечественных площадках приходится задумываться о смысле платить за неработающие функции.

ULTRAVIOLENC3:
I made a PR fixing this issue, you may test it yourself.

[...]

PR #2768, opened 2026-06-25
fix(windows): create DNS WFP sublayer for app split tunneling

ULTRAVIOLENC3:
## Summary

A user reported that app-based split tunneling still failed in #2586 even after executable paths were converted correctly.

Their service log showed:

`Failed to set Config err code 2150760457`

This is `FWP_E_ALREADY_EXISTS`, and the failure happened while applying WFP split-tunnel configuration.

## Changes

The packaged split-tunnel driver expects both WFP sublayers to exist:

- baseline sublayer
- DNS sublayer

Amnezia only created the baseline sublayer. This change makes Windows firewall initialization create/check both sublayers.

## Scope

Windows-only change.

This is separate from the executable path conversion fix (https://github.com/amnezia-vpn/amnezia-client/pull/2691). It addresses the WFP configuration failure reported in #2586.


ygurov (approved):

ygurov:
Hi @ULTRAVIOLENC3,
Thanks for the contribution!

PR #2783, opened 2026-06-28
apple arm64 build support

rlex:
This is initial version of arm64 build support.
I was able to build it locally, and it works (checked with few awg configurations), but i wasn't able to verify github ci pipeline and amnezia premium stuff - since it requires some secrets i don't have access too.

Overall pretty trivial change except QT part - since it already compiles properly, just needed several flags here and there.


rlex:
Oh and notarization / signing part as well, since i don't have apple dev account - but it should be handled by your github actions just the same.

rlex:
(i also noticed that native build connects _much_ faster than rosetta one)

rlex:
Well, looks like it was implemented differently by Amnezia team - they're building universal pkg now. Closing this.

PR #2799, opened 2026-07-03
Adapt openvpn installation to conan

cd-amn:


service/server/CMakeLists.txt
```diff
@@ -356,6 +356,8 @@ if(WIN32)
 
     find_package(wintun REQUIRED)
     list(APPEND CONAN_BINS $<TARGET_FILE:zx2c4::wintun>)
+
+    list(APPEND CONAN_BINS $<TARGET_FILE:OpenSSL::SSL> $<TARGET_FILE:OpenSSL::Crypto>)
```
ygurov, line 360: Are there any binaries in the packages?

service/server/CMakeLists.txt
```diff
@@ -376,8 +378,6 @@ add_custom_command(TARGET ${PROJECT} POST_BUILD
         "$<TARGET_FILE_DIR:${PROJECT}>"
 )
 if(WIN32)
-    # using PERMISSIONS on Windows appends read-only flag
```
ygurov, line 379: I think the comment describes the corner case I have encountered in here, so the next person not to rework everything :D
Let's leave it here?

ygurov (changes requested):

ygurov (approved):

PR #2820, opened 2026-07-08
Fix config import parsing and VPN route edge cases

ShiroKSH:
## Summary

- Accept WireGuard native configs that use `Key=Value` formatting and split `AllowedIPs` regardless of whitespace around commas.
- Reject malformed Xray configs that only contain `inbounds`/`outbounds` text but are not valid JSON objects with those arrays.
- Make legacy WG/AWG native config parsing tolerate `=` formatting without crashing on missing values.
- Initialize VPN connection/error state and fix Linux route/killswitch edge cases.

## Root cause

Several import and route paths assumed narrow formatting or uninitialized state: WireGuard parsers required literal `" = "`, Xray import did substring detection before JSON validation, CIDR mask generation accepted invalid prefixes, Linux block-net rules used the wrong flag, and `deleteTun()` inverted the netlink send result.

## Validation

- `git diff --check origin/dev...HEAD`
- `cmake -S . -B %TEMP%/amnezia-client-cmake-verify -DPREBUILTS_ONLY=ON` was attempted with Conan available in a temp venv. Configure reached Conan package resolution/build and then failed in the existing Windows OpenVPN recipe with `No message compiler found`, before project compilation.

### DIFF
--- a/client/core/controllers/selfhosted/importController.cpp
+++ b/client/core/controllers/selfhosted/importController.cpp
@@ -511,9 +511,10 @@ QJsonObject ImportController::extractWireGuardConfig(const QString &data, Config
         if (trimmedLine.startsWith("[") && trimmedLine.endsWith("]")) {
             continue;
         } else {
-            QStringList parts = trimmedLine.split(" = ");
-            if (parts.count() == 2) {
-                configMap[parts.at(0).trimmed()] = parts.at(1).trimmed();
+            const qsizetype separatorIndex = trimmedLine.indexOf('=');
+            if (separatorIndex > 0) {
+                configMap[trimmedLine.left(separatorIndex).trimmed()] =
+                        trimmedLine.mid(separatorIndex + 1).trimmed();
             }
         }
     }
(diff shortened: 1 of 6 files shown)


ShiroKSH:
@vkamn - could you take a look at the config parsing and VPN route fixes in this PR when you have a chance, or route it to the right reviewer? I can make any requested changes or add tests. Thanks!

ShiroKSH:
@vkamn Friendly reminder: we'd appreciate it if you could take a look at this pull request when you have a chance.

ShiroKSH:
Hi maintainers, this PR has been open for over two weeks. When you have a moment, I would appreciate a review. Thank you.

ShiroKSH:
Rebased onto current `dev` in 487a5c3 and b93a495.

The conflict in `client/vpnConnection.cpp:445` keeps exact key matching for `PersistentKeepalive`, preserves embedded `=` characters, and trims the parsed value. `networkUtilities.cpp:90` also avoids the undefined 32-bit shift for a `/0` route.

ygurov (approved):

PR #2852, opened 2026-07-20
fix: ipc input validation

NickVs2015:
fix ipc input validation  during test

### DIFF
--- a/client/platforms/linux/daemon/linuxfirewall.cpp
+++ b/client/platforms/linux/daemon/linuxfirewall.cpp
@@ -454,16 +454,33 @@ void LinuxFirewall::updateDNSServers(const QStringList& servers)
     static QStringList existingServers {};
 
     existingServers = servers;
-    execute(QStringLiteral("iptables -F %1.320.allowDNS").arg(kAnchorName));
-    for (const QString& rule : getDNSRules(servers))
-        execute(QStringLiteral("iptables -A %1.320.allowDNS %2").arg(kAnchorName, rule));
+    const QString chain = QStringLiteral("%1.320.allowDNS").arg(kAnchorName);
+    executeIptables(QStringLiteral("iptables"), {QStringLiteral("-F"), chain});
+    const QStringList ifaces = {
+        QStringLiteral("amn0+"), QStringLiteral("tun0+"), QStringLiteral("tun2+")
+    };
+    for (const QString& server : servers) {
+        for (const QString& iface : ifaces) {
+            executeIptables(QStringLiteral("iptables"),
+                {QStringLiteral("-A"), chain, QStringLiteral("-o"), iface,
+                 QStringLiteral("-d"), server, QStringLiteral("-p"), QStringLiteral("udp"),
+                 QStringLiteral("--dport"), QStringLiteral("53"), QStringLiteral("-j"), QStringLiteral("ACCEPT")});
+            executeIptables(QStringLiteral("iptables"),
+                {QStringLiteral("-A"), chain, QStringLiteral("-o"), iface,
+                 QStringLiteral("-d"), server, QStringLiteral("-p"), QStringLiteral("tcp"),
+                 QStringLiteral("--dport"), QStringLiteral("53"), QStringLiteral("-j"), QStringLiteral("ACCEPT")});
+        }
+    }
 }
 
 void LinuxFirewall::updateAllowNets(const QStringList& servers)
 {
-    execute(QStringLiteral("iptables -F %1.110.allowNets").arg(kAnchorName));
-    for (const QString& rule : getAllowRule(servers))
-        execute(QStringLiteral("iptables -A %1.110.allowNets %2").arg(kAnchorName, rule));
+    const QString chain = QStringLiteral("%1.110.allowNets").arg(kAnchorName);
+    executeIptables(QStringLiteral("iptables"), {QStringLiteral("-F"), chain});
+    for (const QString& server : servers)
+        executeIptables(QStringLiteral("iptables"),
+            {QStringLiteral("-A"), chain, QStringLiteral("-d"), server,
+             QStringLiteral("-j"), QStringLiteral("ACCEPT")});
 }
 
 void LinuxFirewall::updateBlockNets(const QStringList& servers)
@@ -471,9 +488,12 @@ void LinuxFirewall::updateBlockNets(const QStringList& servers)
     static QStringList existingServers {};
 
     existingServers = servers;
-    execute(QStringLiteral("iptables -F %1.120.blockNets").arg(kAnchorName));
-    for (const QString& rule : getBlockRule(servers))
-        execute(QStringLiteral("iptables -A %1.120.blockNets %2").arg(kAnchorName, rule));
+    const QString chain = QStringLiteral("%1.120.blockNets").arg(kAnchorName);
+    executeIptables(QStringLiteral("iptables"), {QStringLiteral("-F"), chain});
+    for (const QString& server : servers)
+        executeIptables(QStringLiteral("iptables"),
+            {QStringLiteral("-A"), chain, QStringLiteral("-d"), server,
+             QStringLiteral("-j"), QStringLiteral("REJECT")});
 }
 
 int waitForExitCode(QProcess& process)
@@ -499,6 +519,24 @@ int LinuxFirewall::execute(const QString &command, bool ignoreErrors)
         logger.warning()  << "(" << exitCode << ") $ " << command;
     else if (false)
         logger.debug() << "(" << exitCode << ") $ " << command;
+    if (!out.isEmpty())
+        logger.info() << out;
+    if (!err.isEmpty())
+        logger.warning() << err;
+    return exitCode;
+}
+
+int LinuxFirewall::executeIptables(const QString &program, const QStringList &args, bool ignoreErrors)
+{
+    QProcess p;
+    p.start(program, args, QProcess::ReadOnly);
+    p.closeWriteChannel();
+
+    int exitCode = waitForExitCode(p);
+    auto out = p.readAllStandardOutput().trimmed();
+    auto err = p.readAllStandardError().trimmed();
+    if ((exitCode != 0 || !err.isEmpty()) && !ignoreErrors)
+        logger.warning() << "(" << exitCode << ") $ " << program << args.join(QLatin1Char(' '));
     if (!out.isEmpty())
         logger.info() << out;
     if (!err.isEmpty())
--- a/client/platforms/linux/daemon/linuxfirewall.cpp
+++ b/client/platforms/linux/daemon/linuxfirewall.cpp
@@ -33,6 +33,7 @@
 #include "linuxfirewall.h"
 #include "logger.h"
 #include "xray_defs.h"
+#include <QFileInfo>
 #include <QProcess>
 
 #define BRAND_CODE "amn"
@@ -109,7 +110,7 @@ int LinuxFirewall::linkChain(LinuxFirewall::IPVersion ip, const QString& chain, 
         //    (we can't safely delete all rules at once since rule numbers change)
         // TODO: occasionally this script results in warnings in logs "Bad rule (does a matching rule exist in the chain?)" - this happens when
         // the e.g OUTPUT chain is empty but this script attempts to delete things from it anyway. It doesn't cause any problems, but we should still fix at some point..
-        return execute(QStringLiteral("if ! %1 -L %2 -n --line-numbers -t %4 2> /dev/null | awk 'int($1) == 1 && $2 == \"%3\" { found=1 } END { if(found==1) { exit 0 } else { exit 1 } }' ; then %1 -I %2 -j %3 -t %4 && %1 -L %2 -n --line-numbers -t %4 2> /dev/null | awk 'int($1) > 1 && $2 == \"%3\" { print $1; exit }' | xargs %1 -t %4 -D %2 ; fi").arg(cmd, parent, chain, tableName));
+        return execute(QStringLiteral("if ! %1 -L %2 -n --line-numbers -t %4 2> /dev/null | awk 'int($1) == 1 && $2 == \"%3\" { found=1 } END { if(found==1) { exit 0 } else { exit 1 } }' ; then %1 -I %2 -j %3 -t %4 && %1 -L %2 -n --line-numbers -t %4 2> /dev/null | awk 'int($1) > 1 && $2 == \"%3\" { print $1; exit }' | xargs -r %1 -t %4 -D %2 ; fi").arg(cmd, parent, chain, tableName));
     }
     else
         return execute(QStringLiteral("if ! %1 -C %2 -j %3 -t %4 2> /dev/null ; then %1 -A %2 -j %3 -t %4; fi").arg(cmd, parent, chain, tableName));
@@ -290,6 +291,8 @@ void LinuxFirewall::install()
     installAnchor(IPv4, QStringLiteral("120.blockNets"), {});
 
     installAnchor(IPv4, QStringLiteral("110.allowNets"), {});
+
+    installAnchor(Both, QStringLiteral("400.allowPIA"), {});
 
     installAnchor(Both, QStringLiteral("100.blockAll"), {
                                                             QStringLiteral("-j REJECT"),
@@ -546,8 +549,19 @@ int LinuxFirewall::executeIptables(const QString &program, const QStringList &ar
 
 void LinuxFirewall::setupTrafficSplitting()
 {
+    const QString cgroupBase = QStringLiteral("/sys/fs/cgroup/net_cls");
+    if (!QFileInfo::exists(cgroupBase)) {
+        logger.warning() << "net_cls cgroup v1 not available, traffic splitting disabled";
+        return;
+    }
+
+    execute(QStringLiteral(
+        "if ! grep -qE '^[0-9]+[[:space:]]+%1$' /etc/iproute2/rt_tables 2>/dev/null ; then "
+        "echo '200 %1' >> /etc/iproute2/rt_tables ; fi"
+    ).arg(kRtableName));
+
     auto cGroupDir = "/sys/fs/cgroup/net_cls/" BRAND_CODE "vpnexclusions/";
-    logger.info() << "Should be setting up cgroup in" << cGroupDir << "for traffic splitting";
+    logger.info() << "Setting up cgroup in" << cGroupDir << "for traffic splitting";
     execute(QStringLiteral("if [ ! -d %1 ] ; then mkdir %1 ; sleep 0.1 ; echo %2 > %1/net_cls.classid ; fi").arg(cGroupDir).arg(kCGroupId));
     // Set a rule with priority 100 (lower priority than local but higher than main/default, 0 is highest priority)
     execute(QStringLiteral("if ! ip rule list | grep -q %1 ; then ip rule add from all fwmark %1 lookup %2 pri 100 ; fi").arg(kPacketTag, kRtableName));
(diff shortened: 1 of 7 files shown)


client/platforms/linux/daemon/linuxfirewall.cpp
```diff
@@ -454,26 +454,46 @@ void LinuxFirewall::updateDNSServers(const QStringList& servers)
     static QStringList existingServers {};
 
     existingServers = servers;
-    execute(QStringLiteral("iptables -F %1.320.allowDNS").arg(kAnchorName));
-    for (const QString& rule : getDNSRules(servers))
```
ygurov, line 458: Please delete unused `getDNSRules`

client/platforms/linux/daemon/linuxfirewall.cpp
```diff
@@ -454,26 +454,46 @@ void LinuxFirewall::updateDNSServers(const QStringList& servers)
...
+                 QStringLiteral("-d"), server, QStringLiteral("-p"), QStringLiteral("tcp"),
+                 QStringLiteral("--dport"), QStringLiteral("53"), QStringLiteral("-j"), QStringLiteral("ACCEPT")});
+        }
+    }
 }
 
 void LinuxFirewall::updateAllowNets(const QStringList& servers)
 {
-    execute(QStringLiteral("iptables -F %1.110.allowNets").arg(kAnchorName));
-    for (const QString& rule : getAllowRule(servers))
```
ygurov, line 465: Please get rid of unused `getAllowRule`

client/platforms/linux/daemon/linuxfirewall.cpp
```diff
@@ -454,26 +454,46 @@ void LinuxFirewall::updateDNSServers(const QStringList& servers)
...
+             QStringLiteral("-j"), QStringLiteral("ACCEPT")});
 }
 
 void LinuxFirewall::updateBlockNets(const QStringList& servers)
 {
     static QStringList existingServers {};
 
     existingServers = servers;
-    execute(QStringLiteral("iptables -F %1.120.blockNets").arg(kAnchorName));
-    for (const QString& rule : getBlockRule(servers))
```
ygurov, line 475: And from this `getBlockRule` one

ipc/ipc.h
```diff
@@ -15,7 +17,8 @@ enum PermittedProcess {
     OpenVPN,
     Wireguard,
     Tun2Socks,
-    CertUtil
+    CertUtil,
+    _Count
```
ygurov: This exposes _Count to the whole ::amnezia namespace. Consider calling it something like `PermittedProcessCount`, or make PermittedProcess a `enum class`

ipc/ipc.h
```diff
@@ -57,16 +60,56 @@ inline QStringList sanitizeArguments(PermittedProcess proc, const QStringList &a
     QList<Validator> positionalArgs;
 
     switch (proc) {
+    case OpenVPN: {
+        static const QSet<QString> blocked = {
```
ygurov: Blacklist strategy here is weaker than whitelisting IMO.
We have a very strict case of usage of OpenVPN, and we decide which args to call from the client application.
Let's put here only the fields which are actually used in client. It would be also useful to reuse the existing validation code

ipc/ipc.h
```diff
@@ -57,16 +60,56 @@ inline QStringList sanitizeArguments(PermittedProcess proc, const QStringList &a
...
+                qWarning() << "IPC: blocked OpenVPN argument:" << args[i];
+                ++i; // skip following value
+                continue;
+            }
+            out << args[i];
+        }
+        return out;
+    }
+    case Wireguard: {
+        static const QRegularExpression hookRe(
```
ygurov: Same here. Use whitelisting strategy instead of blacklisting

service/server/localserver.cpp
```diff
@@ -17,6 +17,35 @@
...
 
+#ifdef Q_OS_LINUX
+#include <sys/socket.h>
+#include <sys/types.h>
+#include <unistd.h>
+
+uid_t g_allowedUid = static_cast<uid_t>(-1);
+bool g_allowedUidSet = false;
+
+static bool checkPeerCredentials(QLocalSocket *socket) {
```
ygurov: What's the point of this feature?
Theoretically, when the main app crashes - this thingy on the server keeps the previous UID. So it will not pass any other connections.
Client does not have any handles of restarting the service manually - so it will be stuck till manual restart or system reset

service/server/killswitch.cpp
```diff
@@ -11,6 +12,37 @@
...
+    }
+    if (value.contains(QLatin1Char('/'))) {
+        bool ok;
+        int prefix = value.split(QLatin1Char('/'))[1].toInt(&ok);
+        if (!ok || prefix < 0 || prefix > 32) return false;
+    }
+    return true;
+}
+
+static QStringList filterIpList(const QStringList &values) {
```
ygurov: I think filtering-out invalid values is a bugprone way in here. Let's just print an error and return if any of them is not valid.
Modernize it: use `std::all_of` for this task, or similar

ygurov (changes requested):

client/platforms/linux/daemon/linuxfirewall.cpp
NickVs2015: ok, solved it

client/platforms/linux/daemon/linuxfirewall.cpp
NickVs2015: solved

client/platforms/linux/daemon/linuxfirewall.cpp
NickVs2015: solved

ipc/ipc.h
NickVs2015: ok, this lines was changed to PermittedProcessCount like you advise

ipc/ipc.h
NickVs2015: switch to whitelisting strategy

service/server/localserver.cpp
NickVs2015: in that condition ,scenario with different UIDs is practically impossible in real usage (the service runs as root, the client runs as the user), and a client crash does not change its UID on restart — the same user will reconnect with the same UID, so the service will not get stuck. But i agree that it can be discussed

service/server/killswitch.cpp
NickVs2015: ok, change it

service/server/localserver.cpp
ygurov: Ah, nevermind, I thought I saw pid here.

What's the point limiting multiple UIDs? I don't think that's a vector of attack IMO. Moreover, it uses the 1st one as "allowed", which opens a new exploits window - an attacker could try to connect to the IPC before the actual client.

service/server/killswitch.cpp
```diff
@@ -11,6 +13,34 @@
...
+        int prefix = value.split(QLatin1Char('/'))[1].toInt(&ok);
+        if (!ok || prefix < 0 || prefix > 32) return false;
+    }
+    return true;
+}
+
+static QStringList validateIpList(const QStringList &values) {
+    if (!std::all_of(values.cbegin(), values.cend(), isValidIpOrCidr)) {
+        qWarning() << "IPC: IP list contains invalid value, rejecting entire list";
+        return {};
```
ygurov: That's weird reporting the warning which is actually critical, and do nothing in this case.
Why don't we check all the fields at the start of specific calls?

ipc/ipc.h
```diff
@@ -57,16 +60,46 @@ inline QStringList sanitizeArguments(PermittedProcess proc, const QStringList &a
     QList<Validator> positionalArgs;
 
     switch (proc) {
+    case OpenVPN: {
+        // Whitelist only args actually used by the client:
+        // --config <path>, --management <host> <port>, --management-client
+        QStringList out;
+        for (int i = 0; i < args.size(); ++i) {
```
ygurov: There is a mechanism of validation, which is already implemented. Why do we need the same thingy twice?
Check tun2socks implementation below

ipc/ipcserverprocess.cpp
```diff
@@ -77,6 +77,11 @@ void IpcServerProcess::setProcessChannelMode(QProcess::ProcessChannelMode mode)
 
 void IpcServerProcess::setProgram(int programId)
 {
+    if (programId <= static_cast<int>(amnezia::PermittedProcess::Invalid) ||
```
ygurov: I think these checks are making the code to look complex without a reason, and they do nothing actually.
1. If the error could happen returned - tell the caller about it. There is no point of logging the error without the actual error, program does not know what to do with it
2. This call is used only by GUI application (and attackers maybe), so there is no check required IMO. You can also do asserts in here instead, to highlight an error for the developer

service/server/killswitch.cpp
NickVs2015: moved it

ipc/ipc.h
NickVs2015: reused existed code

ipc/ipcserverprocess.cpp
NickVs2015: add assert

service/server/localserver.cpp
NickVs2015: revert it, but it potencial risk which was describe in independed valnurable research

ipc/ipc.h
```diff
@@ -57,16 +60,40 @@ inline QStringList sanitizeArguments(PermittedProcess proc, const QStringList &a
...
+        positionalArgs.append([](const QString& v) {
+            bool ok;
+            int port = v.toInt(&ok);
+            return ok && port > 0 && port <= 65535;
+        });
+        break;
+    }
+    case Wireguard: {
+        // Whitelist only subcommand + config file path (wg-quick up/down <conf>)
+        if (args.size() == 2) {
```
ygurov: 1. Validator does not really care which args sequence does user provide - it's the problem of wg-quick itself. This is to be handled by wg-quick side.
2. We do not use Wireguard in our app, and (I think) do not provide any execs of those names. Could be a vector of attack

So, I would recommend neither:
- To remove WireGuard out of here completely
- Or to use the same positional/named args in here as well as for Tun2Socks, for example

ipc/ipc.h
```diff
@@ -57,16 +60,40 @@ inline QStringList sanitizeArguments(PermittedProcess proc, const QStringList &a
     QList<Validator> positionalArgs;
 
     switch (proc) {
+    case OpenVPN: {
+        // Whitelist only args actually used by the client:
```
ygurov: Please keep the code clean from the slope comments. They do not provide any information - it is written below

ipc/ipcserver.cpp
```diff
@@ -22,6 +22,27 @@
...
 #endif
 
+#ifdef Q_OS_LINUX
+#include <sys/socket.h>
+#include <sys/types.h>
+
+extern uid_t g_allowedUid;
+extern bool g_allowedUidSet;
+
+static bool checkPrivPeerCredentials(QLocalSocket *socket) {
```
ygurov: Same question as for the WireGuard's localServer daemon. This feature is doubtful

service/server/killswitch.cpp
```diff
@@ -166,7 +189,15 @@ bool KillSwitch::disableAllTraffic() {
 
 bool KillSwitch::resetAllowedRange(const QStringList &ranges) {
 
+#ifdef Q_OS_LINUX
+    if (!std::all_of(ranges.cbegin(), ranges.cend(), isValidIpOrCidr)) {
```
ygurov, line 189: Are those ranges used only on Linux? Do other OSes need the same functionality?

service/server/killswitch.cpp
```diff
@@ -328,23 +370,36 @@ bool KillSwitch::enableKillSwitch(const QJsonObject &configStr, int vpnAdapterIn
...
     dnsServers.append("127.0.0.53");
-    
+
+
     for (auto dns : configStr.value(amnezia::configKey::allowedDnsServers).toArray()) {
         if (!dns.isString()) {
             break;
         }
-        dnsServers.append(dns.toString());
+        const QString dnsStr = dns.toString();
```
ygurov: I would recommend to implement this check at the top of the method (basically, near the ranges check), so you would not repeat yourself for each OSes specifically

service/server/localserver.cpp
```diff
@@ -35,7 +36,8 @@ LocalServer::LocalServer(QObject *parent) : QObject(parent),
 
     QObject::connect(m_server.data(), &QLocalServer::newConnection, this, [this]() {
         qDebug() << "LocalServer new connection";
-        m_serverNode.addHostSideConnection(m_server->nextPendingConnection());
+        QLocalSocket *conn = m_server->nextPendingConnection();
```
ygurov: What's the difference? Why do we change this

NickVs2015: resolve it

ipc/ipcserver.cpp
NickVs2015: removed

service/server/killswitch.cpp
NickVs2015: add other os

ipc/ipc.h
NickVs2015: done

ipc/ipc.h
NickVs2015: removed it

service/server/killswitch.cpp
NickVs2015: done

ipc/ipcserver.cpp
```diff
@@ -48,8 +48,9 @@ int IpcServer::createPrivilegedProcess()
     // Make sure any connections are handed to QtRO
     QObject::connect(pd.localServer.data(), &QLocalServer::newConnection, this, [pd]() {
         qDebug() << "IpcServer new connection";
+        QLocalSocket *conn = pd.localServer->nextPendingConnection();
```
ygurov: why do we need this change? what's the difference?

service/server/killswitch.cpp
ygurov: Why don't we do the filtering for everybody who runs the call, despite the OS used? 
I think wrong ranges could be on any OS. I would suggest to drop `#if` statement completely

service/server/killswitch.cpp
```diff
@@ -189,6 +220,12 @@ bool KillSwitch::resetAllowedRange(const QStringList &ranges) {
 }
 
 bool KillSwitch::addAllowedRange(const QStringList &ranges) {
+#if defined(Q_OS_LINUX) || defined(Q_OS_MACOS)
```
ygurov: Same here. I would suggest to drop `#if` statement completely

service/server/killswitch.cpp
```diff
@@ -286,6 +323,15 @@ bool KillSwitch::enableKillSwitch(const QJsonObject &configStr, int vpnAdapterIn
...
     QStringList allownets;
     QStringList blocknets;
+    QStringList allowedDnsServers;
+    for (const QJsonValue &dns : configStr.value(amnezia::configKey::allowedDnsServers).toArray()) {
+        if (!dns.isString()) break;
+        const QString dnsStr = dns.toString();
+        if (isValidIpOrCidr(dnsStr))
+            allowedDnsServers.append(dnsStr);
+        else if (!dnsStr.isEmpty())
+            qWarning() << "IPC: rejected invalid allowedDnsServer:" << dnsStr;
```
ygurov, line 336: Why don't we do the same check at the top? I would rather prefer returning false from the call then having just a log statement in the logs nobody cares about

service/server/killswitch.cpp
```diff
@@ -317,6 +363,11 @@ bool KillSwitch::enableKillSwitch(const QJsonObject &configStr, int vpnAdapterIn
     LinuxFirewall::setAnchorEnabled(LinuxFirewall::Both, QStringLiteral("000.allowLoopback"), true);
     LinuxFirewall::setAnchorEnabled(LinuxFirewall::Both, QStringLiteral("100.blockAll"), blockAll);
     LinuxFirewall::setAnchorEnabled(LinuxFirewall::IPv4, QStringLiteral("110.allowNets"), allowNets);
+    if (!std::all_of(allownets.cbegin(), allownets.cend(), isValidIpOrCidr) ||
+        !std::all_of(blocknets.cbegin(), blocknets.cend(), isValidIpOrCidr)) {
+        qCritical() << "IPC: invalid IP/CIDR in allownets/blocknets, rejecting enableKillSwitch";
+        return false;
```
ygurov: I would suggest doing this check at the top not to have a situation with semi-initialized killswitch due to wrong `blocknets` field

service/server/killswitch.cpp
```diff
@@ -328,23 +379,25 @@ bool KillSwitch::enableKillSwitch(const QJsonObject &configStr, int vpnAdapterIn
...
 
     // We don't use secondary DNS if primary DNS is AmneziaDNS
-    if (!configStr.value(amnezia::configKey::dns1).toString().contains(amnezia::protocols::dns::amneziaDnsIp)) {
-        dnsServers.append(configStr.value(amnezia::configKey::dns2).toString());
+    if (!dns1.contains(amnezia::protocols::dns::amneziaDnsIp)) {
+        const QString dns2 = configStr.value(amnezia::configKey::dns2).toString();
+        if (isValidIpOrCidr(dns2))
+            dnsServers.append(dns2);
+        else if (!dns2.isEmpty())
+            qWarning() << "IPC: rejected invalid dns2:" << dns2;
```
ygurov: Same here. Why don't we put this checks at the top, and return false?

service/server/killswitch.cpp
```diff
@@ -357,6 +410,11 @@ bool KillSwitch::enableKillSwitch(const QJsonObject &configStr, int vpnAdapterIn
         MacOSFirewall::install();
 
     MacOSFirewall::ensureRootAnchorPriority();
+    if (!std::all_of(allownets.cbegin(), allownets.cend(), isValidIpOrCidr) ||
```
ygurov: Same here. I would put this one to the top and return immediately after the call starts

service/server/killswitch.cpp
```diff
@@ -370,20 +428,23 @@ bool KillSwitch::enableKillSwitch(const QJsonObject &configStr, int vpnAdapterIn
...
 
     // We don't use secondary DNS if primary DNS is AmneziaDNS
-    if (!configStr.value(amnezia::configKey::dns1).toString().contains(amnezia::protocols::dns::amneziaDnsIp)) {
-        dnsServers.append(configStr.value(amnezia::configKey::dns2).toString());
+    if (!dns1.contains(amnezia::protocols::dns::amneziaDnsIp)) {
+        const QString dns2 = configStr.value(amnezia::configKey::dns2).toString();
+        if (isValidIpOrCidr(dns2))
+            dnsServers.append(dns2);
+        else if (!dns2.isEmpty())
+            qWarning() << "IPC: rejected invalid dns2:" << dns2;
```
ygurov: And here. Check comments above

ygurov (changes requested):

service/server/killswitch.cpp
NickVs2015: removed

service/server/killswitch.cpp
NickVs2015: removed condition

service/server/killswitch.cpp
NickVs2015: moved

service/server/killswitch.cpp
NickVs2015: moved

service/server/killswitch.cpp
NickVs2015: moved

service/server/killswitch.cpp
NickVs2015: moved

ygurov (approved):

PR #2865, opened 2026-07-23
fix: add 16kb page size support to amnezia-libxray

NickVs2015:
add support 16 kb to amnezia-libxray

### DIFF
--- a/recipes/amnezia-libxray/conanfile.py
+++ b/recipes/amnezia-libxray/conanfile.py
@@ -1,5 +1,5 @@
 from conan import ConanFile
-from conan.tools.files import get, copy
+from conan.tools.files import get, copy, apply_conandata_patches, export_conandata_patches
 from conan.tools.layout import basic_layout
 from conan.errors import ConanInvalidConfiguration
 from conan.tools.env import Environment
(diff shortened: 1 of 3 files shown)


recipes/amnezia-libxray/conanfile.py
```diff
@@ -47,6 +47,11 @@ def _patch_sources(self):
         build_path = os.path.join(self.build_folder, "build.sh")
         build_stat = os.stat(build_path)
         os.chmod(build_path, build_stat.st_mode | stat.S_IEXEC)
+        replace_in_file(self,
+            build_path,
+            '-ldflags="-w -s -buildid= -checklinkname=0"',
+            '-ldflags="-w -s -buildid= -checklinkname=0 -extldflags=-Wl,-z,max-page-size=16384"',
```
ygurov: This change could be done by using a patching mechanism Conan does have.
We did use it in there:
https://github.com/amnezia-vpn/amnezia-client/blob/431ece7fff3b3e5a5042fb3cb5723df79b1de784/recipes/openvpn/conandata.yml#L3

I would suggest doing that instead of a plain replace - git will protect us from bad changes in that script

ygurov (changes requested):

ygurov (approved):

Issue #1324, opened 2024-12-29
Linux ARM64 Build failed 


q4a:
There are already many single board computers like Raspberry Pi.
I tried to build from source on Rock 5B (ARM64, rk3588), but it didn't work. Here is the log:
```
$ uname -a
Linux rock-5b 6.12.1-edge-rockchip-rk3588 #1 SMP PREEMPT Fri Nov 22 14:30:26 UTC 2024 aarch64 aarch64 aarch64 GNU/Linux
$ cat /etc/lsb-release 
DISTRIB_ID=Ubuntu                                                                                                                                                                      
DISTRIB_RELEASE=24.10                                                                                                                                                                  
DISTRIB_CODENAME=oracular                                                                                                                                                              
DISTRIB_DESCRIPTION="Ubuntu 24.10"                                                                                                                                                     
$ echo $QT_BIN_DIR
/usr/lib/qt6/bin                                                                                                                                                                       
$ ./deploy/build_linux.sh 
Build script started ...                                                                                                                                                               
Project dir: /media/q/arm/git/amnezia-client                                                                                                                                           
Build dir: /media/q/arm/git/amnezia-client/deploy/build                                                                                                                                
Using Qt in /usr/lib/qt6/bin                                                                                                                                                           
cmake version 3.30.3                                                                                                                                                                   
                                                                                                                                                                                       
CMake suite maintained and supported by Kitware (kitware.com/cmake).
Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/usr/libexec/gcc/aarch64-linux-gnu/14/lto-wrapper
OFFLOAD_TARGET_NAMES=nvptx-none
OFFLOAD_TARGET_DEFAULT=1
Target: aarch64-linux-gnu
[...]

PR #2870, opened 2026-07-25
fix: make Conan recipes parse on Python < 3.12

Mezo-oz:
### Problem

Three Conan recipes use f-string syntax that is only valid on **Python 3.12+** ([PEP 701](https://peps.python.org/pep-0701/)):

- `recipes/amnezia-xray-bindings/conanfile.py:115`
- `recipes/tun2socks/conanfile.py:118`
- `recipes/hev-socks5-tunnel/conanfile.py:63-65`

Two things are involved, both illegal before 3.12: reusing the same quote character inside an f-string expression, and using a backslash inside one.

```python
f"BUILD_DIR={build_dir.replace("\\", "/") if self._is_windows else build_dir}"
                              ^ SyntaxError
```

On an older interpreter Conan fails to load the recipe *before resolving a single dependency*, so the build cannot start at all:

```
ERROR: Error loading conanfile at '.../recipes/amnezia-xray-bindings/conanfile.py'
  File ".../conanfile.py", line 115
    f"BUILD_DIR={build_dir.replace("\\", "/") if self._is_windows else build_dir}"
                                     ^
SyntaxError: unexpected character after line continuation character
CMake Error at cmake/conan_provider.cmake:504 (message):
  Conan install failed='1'
```

This is **not platform specific** — it affects any host with Python < 3.12. It is invisible in CI only because `ubuntu-latest` ships Python 3.12+. Debian 12 (bookworm) ships Python 3.11, remains widely deployed, and was the base for Raspberry Pi OS through 2025.

### Fix

Hoist the expressions out of the f-strings, and use single quotes for the nested string literals. Behaviour is identical on every Python version; no functional change.

### Verification

`python3 -m py_compile` on **Python 3.11.2** (aarch64, Debian 12):

| recipe | before | after |
|---|---|---|
| `amnezia-xray-bindings` | SyntaxError | OK |
| `tun2socks` | SyntaxError | OK |
| `hev-socks5-tunnel` | SyntaxError | OK |

With these applied, `deploy/build.sh` proceeds normally and Conan builds the full dependency graph from source — `go/1.26.0`, `openssl/3.6.x`, `libssh/0.11.3@amnezia`, `awg-go/0.2.19`, `openvpn/2.7.0`, `tun2socks/2.6.0`, `amnezia-xray-bindings/1.2.0` and the rest — on `arch=armv8`.

Related to #1324

### DIFF
--- a/recipes/amnezia-xray-bindings/conanfile.py
+++ b/recipes/amnezia-xray-bindings/conanfile.py
@@ -111,8 +111,9 @@ class AmneziaXrayBindings(ConanFile):
                 env.define("CGO_LDFLAGS", " ".join(ldflags))
                 with env.vars(self).apply():
                     at = Autotools(self)
+                    make_build_dir = build_dir.replace("\\", "/") if self._is_windows else build_dir
                     at.make(args=[
-                        f"BUILD_DIR={build_dir.replace("\\", "/") if self._is_windows else build_dir}"
+                        f"BUILD_DIR={make_build_dir}"
                     ])
 
             if is_apple_os(self) and self._is_multiarch:
--- a/recipes/hev-socks5-tunnel/conanfile.py
+++ b/recipes/hev-socks5-tunnel/conanfile.py
@@ -60,9 +60,9 @@ class HevSocks5Tunnel(ConanFile):
             self.run(
                 f"libtool -static -o {lib_path}"
                 f" {lib_path}"
-                f" {os.path.join(self.build_folder, "third-part", "lwip", "bin", "liblwip.a")}"
-                f" {os.path.join(self.build_folder, "third-part", "yaml", "bin", "libyaml.a")}"
-                f" {os.path.join(self.build_folder, "third-part", "hev-task-system", "bin", "libhev-task-system.a")}"
+                f" {os.path.join(self.build_folder, 'third-part', 'lwip', 'bin', 'liblwip.a')}"
+                f" {os.path.join(self.build_folder, 'third-part', 'yaml', 'bin', 'libyaml.a')}"
+                f" {os.path.join(self.build_folder, 'third-part', 'hev-task-system', 'bin', 'libhev-task-system.a')}"
             )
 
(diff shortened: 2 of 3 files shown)


Mezo-oz:
Verified end to end: with this fix, Conan resolves and builds the full dependency graph on Debian 12 with Python 3.11.2.

Environment — Raspberry Pi 5, aarch64, `debian:bookworm` container:

| | |
|---|---|
| Python | 3.11.2 |
| Conan | 2.31.1 |
| CMake | 3.25.1 |
| gcc | 12.2.0 |
| Qt | 6.4.2 (distro) |

All recipes parse, the graph resolves, and the dependencies build from source — OpenSSL, libssh, Go, tun2socks, v2ray-rules-dat and the rest. The project then configures and builds to completion:

```
[262/279] Linking CXX executable client/AmneziaVPN
[279/279] Linking CXX executable service/server/AmneziaVPN-service
```

Without this fix the PEP 701 f-strings raise `SyntaxError` on Python 3.11 before a single dependency is resolved, so nothing downstream is reachable.

Worth noting the practical impact is not limited to older distros generally — it is most acute on ARM64 Linux, where Qt publishes no official desktop binaries and distro Qt is the only option, so the system Python tends to come with it. This build was done while testing exactly that case for #1324.

Mezo-oz:
Still applies cleanly on `dev` as of `7d4f3e0f` — none of the three recipes have changed since this was opened, so the merge is conflict-free and the verification above still stands.

Happy to rebase or split it up if that would help review.

Also fair if you'd rather go the other direction: declaring Python 3.12 the minimum and documenting it is a reasonable call, it just means anyone on Debian 12 or Raspberry Pi OS has to install a newer interpreter before the build will start at all. This is only worth merging if supporting the distro Python is something you want to keep.

ygurov (approved):

Slach:
@Mezo-oz thanks for your effort hope Amnezia VPN client will available on Linux ARM64 in near future

PR #2900, opened 2026-07-29
fix: build and run on distro Qt by dropping 6.5+ API dependencies

Mezo-oz:
The client required Qt 6.9 without ever declaring it, so on older Qt it built cleanly and then failed at runtime. Every version-gated API involved turned out to be avoidable, so this removes the dependencies rather than raising the minimum.

Five blockers, in the order a build hits them:

| API | Requires | Change |
|---|---|---|
| `OUTPUT_SCRIPT` (was `FILENAME_VARIABLE`) | 6.5 | branch on `Qt6_VERSION` |
| `DEPLOY_TOOL_OPTIONS` | 6.7 | passed only on ≥ 6.7 |
| `popupType: Popup.Native` | 6.8 | removed |
| `ContextMenu.menu` attached type | 6.9 | explicit `MouseArea` + `Menu.popup()` |
| `pragma ComponentBehavior: Bound` | 6.5 | removed from two files — see below |

There are **two** deploy-script call sites, not one: `client/CMakeLists.txt` uses `qt_generate_deploy_qml_app_script` (Qt6QmlMacros) and `service/server/CMakeLists.txt` uses `qt_generate_deploy_app_script` (Qt6CoreMacros). They live in different Qt modules and changed signature together.

The keyword sets were read off the Qt 6.4 sources rather than the documentation, which describes only the current signature — Qt 6.4 accepts `TARGET` / `FILENAME_VARIABLE` and no multi-value options at all.

### The context menu

`ContextMenu.menu` is replaced by a `MouseArea` accepting only `Qt.RightButton`, so left press and drag still reach the text control for cursor placement, selection and flicking. Confirmed by hand on a real display: the menu opens on right-click, Cut/Copy enable and disable with the selection, and normal editing is unaffected.

### One decision worth the team's call

`QuestionDrawer.qml` and `SubscriptionExpiredDrawer.qml` carry `pragma ComponentBehavior: Bound`, and `DrawerType2` instantiates their content through an always-active `Loader` — outside its creation context, which Qt 6.4 rejects:

```
Cannot instantiate bound component outside its creation context
```

The `Loader` then reports status `Error` and its `item` stays null, so the drawer opens with no text and no buttons. The confirmation dialogs in the server, protocol and config removal flows cannot be acted on. Isolated reproduction, same code on both versions:

| | Qt 6.4.2 | Qt 6.10.2 |
|---|---|---|
| Loader status | 3 (Error) | 1 (Ready) |
| `item` | null | created |

The pragma can be dropped from these two files without any behaviour change: it was added in b457ef9a together with qualifying every reference (`text: headerText` → `text: root.headerText`), so the files already satisfy what it enforces. Verified identical results on 6.4.2 and 6.10.2 after removal.

That said, it does remove a lint guard — a future edit could introduce unqualified access in those two files without a compile-time error. **Happy to restore the pragma and set the minimum at 6.5 instead if you prefer that trade.** `AwgTextField.qml` keeps its pragma either way, since it is used as a direct type and never instantiated through a `Loader`.

### Declaring the minimum

`find_package(Qt6 ...)` had no version constraint, which is why an unsupported Qt produced a runtime QML failure instead of a configure-time error. It now declares `6.4`, so falling below the supported minimum fails immediately and names both versions:

```
Could not find a configuration file for package "Qt6" that is compatible
with requested version "6.4".
```

6.4.2, 6.8.2 and 6.10.2 were each built and run; 6.5 through 6.7 were not tested, as no distro to hand ships them. The floor is set to the lowest version actually exercised. If you would rather it track the oldest Qt you intend to support, that number is easy to change — the point is that it is declared rather than implicit.

### Verification

Raspberry Pi 5, aarch64. Built and run on each:

| Qt | Source | Build | Runs |
|---|---|---|---|
| 6.4.2 | Debian 12 bookworm | client + service | ✅ |
| 6.8.2 | Debian 13 trixie, native | client + service | ✅ + right-click menu confirmed on a real display |
| 6.10.2 | Ubuntu 26.04 | — | ✅ |

"Runs" means the QML engine loads and the client reaches `Started AmneziaVPN version 5.0.0.2`; before these changes it failed with `QQmlApplicationEngine failed to load component`.

CI is unaffected — the workflows install Qt 6.10.1, which takes the same branches as today.

Related: #1324, where ARM64 Linux users are affected first because Qt publishes no official aarch64 desktop binaries, making distro Qt the only option there.

### DIFF
--- a/client/CMakeLists.txt
+++ b/client/CMakeLists.txt
@@ -38,7 +38,7 @@ if(WIN32 OR (APPLE AND NOT IOS) OR (LINUX AND NOT ANDROID))
     set(PACKAGES ${PACKAGES} Widgets)
 endif()
 
-find_package(Qt6 REQUIRED COMPONENTS ${PACKAGES})
+find_package(Qt6 6.4 REQUIRED COMPONENTS ${PACKAGES})
 
 set(LIBS ${LIBS}
     Qt6::Core Qt6::Gui
@@ -246,12 +246,23 @@ if(WIN32)
     set(deploy_tool_options "--force-openssl --force")
 endif()
 
-qt_generate_deploy_qml_app_script(
-    TARGET ${PROJECT}
-    OUTPUT_SCRIPT QT_DEPLOY_SCRIPT
-    NO_UNSUPPORTED_PLATFORM_ERROR
-    DEPLOY_TOOL_OPTIONS ${deploy_tool_options}
-)
+# DEPLOY_TOOL_OPTIONS was added in Qt 6.7; passing it to an older Qt fails at
+# configure time with an unknown-keyword error. It is empty on every platform
+# except Windows, so older Qt simply omits it and behaves identically.
+if(Qt6_VERSION VERSION_GREATER_EQUAL 6.7)
+    qt_generate_deploy_qml_app_script(
+        TARGET ${PROJECT}
+        OUTPUT_SCRIPT QT_DEPLOY_SCRIPT
+        NO_UNSUPPORTED_PLATFORM_ERROR
+        DEPLOY_TOOL_OPTIONS ${deploy_tool_options}
+    )
+else()
+    qt_generate_deploy_qml_app_script(
+        TARGET ${PROJECT}
+        OUTPUT_SCRIPT QT_DEPLOY_SCRIPT
+        NO_UNSUPPORTED_PLATFORM_ERROR
+    )
+endif()
 install(SCRIPT ${QT_DEPLOY_SCRIPT}
     COMPONENT AmneziaVPN
 )
--- a/client/CMakeLists.txt
+++ b/client/CMakeLists.txt
@@ -246,9 +246,16 @@ if(WIN32)
     set(deploy_tool_options "--force-openssl --force")
 endif()
 
-# DEPLOY_TOOL_OPTIONS was added in Qt 6.7; passing it to an older Qt fails at
-# configure time with an unknown-keyword error. It is empty on every platform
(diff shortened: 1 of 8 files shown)


Mezo-oz:
Rebased onto current `dev`. The conflicts are resolved and it merges cleanly now.

Since #2916 moved long-press through `ContextMenu.onRequested` → `requestNative()`, the original right-click `MouseArea` would have broken iOS. So the fix changed shape: `ContextMenuType.qml` keeps all of the iOS code (`useNativeEditMenu`, `requestNative`) untouched and adds a small `show(position)` helper. The three input types use two `TapHandler`s instead of the attached type: right-click for desktop, long-press for touch. Left press and drag still reach the text item.

**Tested:** built on a RPi (Qt 6.8.2, Debian 13), and I checked the right-click menu (cut, copy, paste) by hand on a display (VcXsrv).

**Not re-tested:** the Qt 6.4.2 / Debian 12 build was verified on 07-28 and hasn't been re-run since this rebase. The rebase didn't touch those deploy-script paths.

**Tradeoff:** dropping the `ContextMenu` attached type means that on Qt 6.10+ iOS, where `useNativeEditMenu` is false, users get the Qt-drawn menu instead of the native one. Dropping `popupType: Popup.Native` does the same on desktop Qt 6.8+. That's a real regression on new Qt in exchange for working at all on older distro Qt. If you'd rather keep the native menus, the alternative is to raise the minimum Qt version to 6.9 and leave `ContextMenu` in place. I'm happy to switch the PR to that.

Issue #2907, opened 2026-07-30
[macOS] Site split tunneling not applied for amnezia-awg (AmneziaWG 1.x) connections, but works for amnezia-awg2


m8ig:
## Summary

On macOS, site-based split tunneling ("All sites except the listed ones") is silently ignored when connected to a server whose container is **`amnezia-awg`**, while the exact same site list is applied correctly when connected to a server whose container is **`amnezia-awg2`**.

The app reports the correct route mode for *both* connections. The difference appears only in what the privileged daemon actually installs into the system routing table: **2100 exclusion host routes** in one case, **zero** in the other.

Both measurements were taken on the same machine, in the same session, on the same network, with the same site list, minutes apart — the only variable was which server was connected.

The same site list works correctly with the **same `amnezia-awg` server** on **iOS**, which suggests this is specific to the macOS route-installation path rather than to the protocol or to the configuration.

## Environment

| | |
|---|---|
| App | AmneziaVPN **5.0.0.5** (`4d28f81`) |
| OS | macOS Tahoe **26.5.2** (arm64) |
| Split tunneling mode | Sites: *All sites except the listed ones* (`VpnAllExceptSites`, route mode `2`) |
| App split tunneling | disabled |
| Site list size | ~2100 IPv4 entries |

Server A — self-hosted, created by the app, container **`amnezia-awg2`** (AmneziaWG 2.0)
Server B — third-party config imported from a `.conf` file, container **`amnezia-awg`** (AmneziaWG 1.x)

## Steps to reproduce

1. Add a self-hosted AmneziaWG 2.0 server (container `amnezia-awg2`).
2. Import a third-party AmneziaWG 1.x `.conf` (container `amnezia-awg`).
3. Enable site split tunneling in *All sites except the listed ones* mode and load a list of IPs.
4. Connect to server A. Inspect the routing table.
5. Connect to server B. Inspect the routing table again.

## Expected

Host routes for every excluded IP are installed via the physical gateway in both cases, so excluded destinations bypass the tunnel.

## Actual

**Connected to server A (`amnezia-awg2`)** — split tunneling works. Excluded IPs get host routes via the physical gateway:

```
0/1                <tunnel>            UScg    utun6
128.0/1            <tunnel>            USc     utun6
default            <gateway>           UGScg   en0
<excluded-ip-1>/32 <gateway>           UGSc    en0
<excluded-ip-2>/32 <gateway>           UGSc    en0
...
```

[...]

Issue #2904, opened 2026-07-30
[Windows] Split tunneling creates /32 routes instead of configured network prefixes in v5.0.0.5


zomoleg:
### Description

After upgrading from **AmneziaVPN 4.8.21** to **5.0.0.5** on Windows, split tunneling by destination networks no longer works.

The same configuration works correctly in **4.8.21**, but in **5.0.0.5** all traffic is routed through the VPN.

### Steps to reproduce

1. Install AmneziaVPN 5.0.0.5 on Windows.
2. Configure split tunneling with several excluded IPv4 networks (for example `5.35.0.0/19`, `31.130.140.0/22`, `37.9.64.0/18`).
3. Connect to the VPN.
4. Run:
CMD:
```
route print
```

or
PowerShell:
```powershell
Get-NetRoute
```

### Expected behavior

Routes should be created with their configured prefix lengths.

Examples:

```
5.35.0.0/19
31.130.140.0/22
37.9.64.0/18
```

Traffic to these networks should bypass the VPN.

### Actual behavior

Every configured network is installed as a **/32 host route** instead of the configured prefix.

Examples:

```
5.35.0.0/19 -> 5.35.0.0/32
31.130.140.0/22 -> 31.130.140.0/32
37.9.64.0/18 -> 37.9.64.0/32
```

As a result, only the network address itself is excluded while all other addresses inside the subnet continue to use the VPN.

### Comparison

Working (4.8.21):

```
5.35.0.0/19
31.130.140.0/22
37.9.64.0/18
77.88.0.0/18
```

Broken (5.0.0.5):

```
5.35.0.0/32
31.130.140.0/32
37.9.64.0/32
77.88.0.0/32
```

This looks like the prefix length is being lost when Windows routes are created.

### Additional information

* Windows version: Windows 11 26H1 28000.2525
* VPN protocol: AmneziaWG (AWG 2.0)
* The same configuration works correctly with AmneziaVPN 4.8.21.
* I attached the outputs of `route print` and `Get-NetRoute` for both versions for comparison.

[Get-NetRoute (4.8.21).txt](https://github.com/user-attachments/files/30539608/Get-NetRoute.4.8.21.txt)
[Get-NetRoute (5.0.0.5).txt](https://github.com/user-attachments/files/30539607/Get-NetRoute.5.0.0.5.txt)
[route print (4.8.21).txt](https://github.com/user-attachments/files/30539610/route.print.4.8.21.txt)
[...]

PR #2912, opened 2026-07-31
fix: keep CIDR prefix length when normalizing split tunneling entries

antib0t:
## Problem

`normalizeHostname()` strips everything after the first slash. It exists to drop the path from a URL, but it runs on every entry, CIDR ranges included, and it runs before validation. So `91.108.4.0/22` gets stored as `91.108.4.0`.

Downstream in `localsocketcontroller.cpp:224-229`, any address without a slash gets a `/32`:

```cpp
} else {
    QJsonObject range;
    range.insert("address", ipRange);
    range.insert("range", 32);
    range.insert("isIpv6", false);
    jsAllowedIPAddesses.append(range);
}
```

A /22 subnet ends up as a host route to its own network address, which nothing talks to, and the rest of the subnet skips the tunnel. Both entry points hit this: `addSite()` at line 68 and `importSitesFromJson()` at line 215. An imported list of ranges lands in settings with every prefix gone.

## Regression

4.8.21.0 had no `normalizeHostname()`. `sitesController.cpp:45-46` stored the hostname as it was typed. The route building code in `localsocketcontroller.cpp` is identical in both versions, so only the input path changed.

## Fix

Return CIDR ranges unchanged and leave URL handling alone.

## Testing

Ran the normalization logic against the regexp for /16, /20, /22, /26 and /32 subnets, a bare IPv4 address, a domain, URLs with paths, and a malformed prefix like `1.2.3.4/999`. Subnets keep their prefix. URLs still lose the path. The malformed one falls back to the bare address, as before.

Verified on macOS 26.5.2 with AmneziaVPN 5.0.0: a list of 2960 subnets imported from iplist.opencck.org was stored with all prefixes stripped, and the routing table showed no exclusion routes beyond the VPN endpoint itself.

## Note for users

Lists imported by an affected build are already stored without prefixes, and upgrading does not bring them back. The list has to be imported again.

Related: #2904, #2907

### DIFF
--- a/client/core/controllers/ipSplitTunnelingController.cpp
+++ b/client/core/controllers/ipSplitTunnelingController.cpp
@@ -144,6 +144,14 @@ QString IpSplitTunnelingController::normalizeHostname(const QString &hostname) c
     normalized.replace("https://", "");
     normalized.replace("http://", "");
     normalized.replace("ftp://", "");
+
+    // A CIDR range is not a URL: the part after the slash is a prefix length,
+    // not a path. Stripping it turns 91.108.4.0/22 into 91.108.4.0, which is
+    // later routed as a /32 host route, so the rest of the subnet misses the tunnel.
+    if (NetworkUtilities::ipAddressWithSubnetRegExp().exactMatch(normalized)) {
+        return normalized;
+    }
+
     normalized = normalized.split("/", Qt::SkipEmptyParts).first();
     return normalized;
 }


vkamn:
Is this issue still present in the current dev branch?

antib0t:
No, not anymore. b575df05 (#2973, Aug 10) makes `normalizeHostname()` return early when the entry matches `ipAddressWithSubnetRegExp()`, so CIDR ranges keep their prefix. I pulled the current dev tree and can't reproduce it there, so I'm closing this as a duplicate.

One thing that might be worth a line in the release notes: lists imported by an affected build are already stored without prefixes, and updating doesn't bring them back. Users have to import the list again.

PR #2923, opened 2026-08-01
fix(i18n): move translations to the renamed contexts

### LINKED ISSUE #2909: Update the localization of the program version 5.0.0.5
Please update the localization for version 5.0.0.5.
After applying patch #2009, some files were renamed and are now not taken into account.
Example: "chore: rename connectionController to connectionUiController".

WkdXeqtr:
After the controller renames a number of translations stayed under context names the code no longer asks for, so they sit in the file but never get applied.

Moving them to the current names brings back 193 translations. Counted as entries whose context and source match what the current sources actually ask for:

| | on dev | after |
| --- | --- | --- |
| ar_EG | 505 | 537 |
| fa_IR | 478 | 510 |
| hi_IN | 447 | 476 |
| my_MM | 541 | 573 |
| uk_UA | 484 | 516 |
| ur_PK | 420 | 449 |
| zh_CN | 425 | 432 |

Russian is not touched here, it is handled in #2980.

### DIFF
--- a/client/translations/amneziavpn_ar_EG.ts
+++ b/client/translations/amneziavpn_ar_EG.ts
@@ -49,24 +49,6 @@
     </message>
 </context>
 <context>
-    <name>ApiConfigsController</name>
-    <message>
-        <location filename="../ui/controllers/api/apiConfigsController.cpp" line="203"/>
-        <source>%1 installed successfully.</source>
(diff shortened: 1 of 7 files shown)


vkamn:
fixed here https://github.com/amnezia-vpn/amnezia-client/pull/3002

Issue #2152, opened 2026-01-23
False disconnected status (actually connected) after PC reboot with Auto connect disabled


morskoyzmey:
**Describe the bug**
Disable Auto connect. Connect to server then reboot PC and client will show it is Disconnected while it is actually connected.

**To Reproduce**
1. Enable Auto start
2. Disable Auto connect
3. Connect to server
4. Reboot PC

**Expected behavior**
App starts but not connects to server.

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Version 4.8.11.4

**Additional context**
Self-hosted server on Ubuntu if it matters

Issue #1258, opened 2024-11-17
Stuck at connecting, no internet after restart, tunnel hanging. See this for fix


Nodaysofff:
**Describe the bug**
App didnt close tunnel, when its quit anyting but its tray icon close button, which makes tunnel to persist after system restart and makes it unable to connect to the server again due to unterminated "tun2socks" process

**To Reproduce**
1. Just launch app as you usually do and use split tunneling
2. Click on 'Connect' and wait for app to establish connection to your server
3. Restart the system
4. See no internet when you`ve restarted, tun2socks and tunnel hanging, sometimes have app stuck on connecting, etc...

**Expected behavior**
App should handle scenario when its closed by the system, since now it closes tunnel only when user closes the app from the tray icon or from inside the app

**Log files**
I`ve described the bug, but in the logs it shows that tun2socks already running and quitting when you stuck at connecting

**Desktop:**
 - OS: Windows 10
 - Version any

**Additional context**
At the moment there is only two ways for you to escape such bug, until its fixed in the code:
1) Before the system shutdown (or anything that will terminate app process) you should close the app manually
2) I`ve made a ps1 script that terminates that process at system startup for you:
- Create a file named "CloseTunnel.ps1" and paste here the following code
```# Define the name of the process to kill
$processName = "tun2socks"

# Get the list of processes matching the name
$process = Get-Process -Name $processName -ErrorAction SilentlyContinue

if ($process) {
    # Terminate the process if it exists
    Stop-Process -Name $processName -Force
    Write-Host "Process '$processName' has been terminated."
} else {
    Write-Host "Process '$processName' is not running."
}
```
- Press WIn+R and paste there "taskschd.msc" to open the Task Scheduler
- Create a new task with any name you want
- Set it to run with the highest privileges by a corresponding checkbox
- Add a new trigger and set it to "At startup."
- Add a new action and set the program/script to your path to PowerShell.exe (usually its C:\Windows\System32\... etc.)
- Add arguments -File "C:\path\to\CloseTunnel.ps1" and replace the path to your just created ps1 file
- Now at startup you`ll see powershell opening and launching that script for a second and getting back you your internet and letting app to autostart correctly

Ehsan-U:
Stuck at connecting.. 
OS: Ubuntu 22

TheGP:
+1 windows 11

farhadsaberi:
+1 Mac os

Ehsan-U:
[...]

PR #2930, opened 2026-08-03
fix: stack overflow while resolving site routes

### LINKED ISSUE #2938: Windows: stack overflow while resolving a large site split-tunneling list
> *This was generated by AI during triage.*

## Environment

- Windows 11, build 26100.8655
- AmneziaVPN 5.0.0.5
- Xray connection using `tun2socks`
- Site split tunneling enabled with a large site list

## Problem

While connecting with a large site split-tunneling list, `AmneziaVPN.exe` exits with Windows exception `0xc00000fd` (stack overflow) as hostnames are re-resolved.

## Reproduction

1. Enable site split tunneling with a large site list. In the observed incident, it produced more than 9,000 route entries.
2. Connect an Xray configuration.
3. Wait while the configured hostnames are resolved.
4. Observe `AmneziaVPN.exe` terminate with `0xc00000fd`.

## Root cause

Each `QHostInfo::lookupHost()` completion enters the callback in `VpnConnection::addSitesRoutes()`. That callback synchronously waits for `flushDns()` through `QRemoteObjectPendingCall::waitForFinished()`.

The synchronous wait processes a nested Qt event loop, which can dispatch the next queued hostname callback before the current callback returns. The observed crash dump contained this logical cycle approximately 66 times:

```text
addSitesRoutes QHostInfo callback
  -> flushDns
  -> QRemoteObjectPendingCall::waitForFinished
  -> nested QEventLoop::exec
  -> next QHostInfo callback
```

The original dump is no longer available, so this is a transcription of the inspected call pattern rather than a raw debugger export.

## Expected behavior

Resolving a large site list must not create nested synchronous event loops or exhaust the GUI thread stack. Resolved IPv4 routes should still be installed, and DNS should be flushed after resolution completes.

## Proposed fix

PR #2930 removes the synchronous wait from the hostname callback, installs each resolved route as its lookup completes, and performs one asynchronous DNS flush after all lookups finish.

## Acceptance criteria

- Connecting with a large site list does not enter a nested `waitForFinished()` loop in `addSitesRoutes()`.
- Each newly resolved IPv4 address is still added to the selected route gateway.
- DNS is flushed once asynchronously after all hostname lookups complete.
- Failures from the asynchronous DNS flush are still logged.

## Out of scope

Recovering or cleaning an orphaned VPN session after any unexpected GUI exit is tracked separately in #2931. This issue covers only the stack-overflow trigger fixed by #2930.

## Related report

#2748 contains the same Windows exception code and reports a remaining interface with broken DNS after the GUI exits, but its primary AmneziaDNS/reinstallation problem is broader and is not claimed to be fixed here.

romanio427:
## Summary

- keep multi-IP site resolution from #2811 (`addVpnSite(..., QStringList)`)
- install each resolved site route as soon as its lookup completes
- track outstanding hostname lookups and flush DNS **once** when they all finish
- flush DNS asynchronously through `QRemoteObjectPendingCallWatcher` (no nested event loop)
- capture `gw` / `ip` by value in the async `routeAddList` lambda (avoids dangling refs from the previous `[&gw, &ip]` capture)

## Problem

`VpnConnection::addSitesRoutes()` synchronously called `QRemoteObjectPendingCall::waitForFinished()` from every asynchronous hostname lookup callback.

`waitForFinished()` processes a nested Qt event loop. With a large site split-tunneling list, another queued `QHostInfo` callback can run before the previous callback returns. The cycle repeats until `AmneziaVPN.exe` terminates with `0xc00000fd` (stack overflow).

A Windows crash dump showed this repeated stack cycle approximately 66 times:

`addSitesRoutes callback -> flushDns -> waitForFinished -> QEventLoop::exec -> next QHostInfo callback`

The old condition also logged failure on successful completion because it used:

`reply.waitForFinished() || !reply.returnValue()`

The asynchronous watcher preserves error reporting without starting nested event loops.

## What changed in this update

Rebased onto current `dev` after #2811 (`fix: enhance split tunneling IP handling`) landed in the same function and produced a merge conflict in `client/vpnConnection.cpp`.

Conflict resolution keeps both sides:

| From #2811 (`dev`) | From this PR |
| --- | --- |
| Resolve **all** IPv4 addresses per site | Do **not** call `waitForFinished()` in the `QHostInfo` callback |
| `addVpnSite(mode, site, newIps)` with `QStringList` | Count remaining lookups; flush DNS **once** at the end |
| Split-tunneling debug log | Async flush via `QRemoteObjectPendingCallWatcher` |
| Flush only when new IPs were installed | Same policy via shared `needFlush` flag |

## Reproduction and dump evidence

Observed environment:

- Windows 11, build 26100.8655
- AmneziaVPN 5.0.0.5
- Xray connection using `tun2socks`
- Kill switch and site split tunneling enabled
- A large site list producing more than 9,000 route entries during the incident

Reproduction:

1. Enable site split tunneling with the large site list.
2. Connect the Xray configuration.
3. While the hostnames are being re-resolved, `AmneziaVPN.exe` exits with `0xc00000fd`.

The dump inspection showed this logical frame cycle repeated approximately 66 times before the stack overflow:

```text
VpnConnection::addSitesRoutes QHostInfo callback
  -> IpcInterfaceReplica::flushDns
  -> QRemoteObjectPendingCall::waitForFinished
  -> nested QEventLoop::exec
  -> next queued QHostInfo callback
  -> VpnConnection::addSitesRoutes QHostInfo callback
```

The original dump is no longer available for attachment, so the cycle above is a transcription of the inspected call pattern rather than a raw debugger export. The patch removes the nested synchronous wait from this callback path.

## Verification notes

- Merge conflicts with `dev` are resolved; PR is mergeable against current `dev`.
- Change is scoped to desktop (`#ifdef AMNEZIA_DESKTOP`); the new include is under the same guard.
- Static review: `flushDns()` is `SLOT(bool flushDns())`, so `QRemoteObjectPendingReply<bool>` + `QRemoteObjectPendingCallWatcher` match the existing Qt Remote Objects IPC stack.
- No local packaged Windows build was produced for this PR. Runtime confirmation still needs: large site list + split tunneling + connect/re-resolve on Windows (the original crash scenario).

## Linked issue

Fixes #2938

## Scope

This PR only fixes the client crash during site-route resolution. Restoring or cleaning up an orphaned desktop VPN session after an unexpected GUI exit is a separate lifecycle concern.

## Related issues

- Related to #2748 — the same Windows stack-overflow exception (`0xc00000fd` in `Qt6Core.dll`) after connecting, followed by the GUI exiting and a network interface remaining active with non-functional DNS. This PR addresses the identified stack-overflow path only; it does not claim to fix the separate AmneziaDNS/reinstallation behavior reported there.
- Related to #1258 — `tun2socks` and its tunnel survive abnormal application or system termination, leaving no internet and preventing a new connection. This is the same orphaned-session aftermath, not the crash root cause fixed by this PR.
- Related to #2152 — after restart, the desktop GUI reports Disconnected while the VPN is still active. This matches the client/service state mismatch observed after our crash.
- Follow-up: #2931 contains the current Windows reproduction and source-level analysis of the orphaned-session lifecycle gap. It is intentionally separate because this PR only fixes the triggering stack overflow.

### DIFF
--- a/client/vpnConnection.cpp
+++ b/client/vpnConnection.cpp
@@ -17,6 +17,7 @@
 #ifdef AMNEZIA_DESKTOP
     #include "core/utils/ipcClient.h"
     #include <core/protocols/wireGuardProtocol.h>
+    #include <QRemoteObjectPendingCallWatcher>
 #endif
 
 #ifdef Q_OS_ANDROID
@@ -230,9 +231,12 @@ void VpnConnection::addSitesRoutes(const QString &gw, amnezia::RouteMode mode)
         iface->routeAddList(gw, ips);
     });
 
+    auto remainingLookups = QSharedPointer<int>::create(sites.size());
+    auto needFlush = QSharedPointer<bool>::create(false);
+
     // re-resolve domains
     for (const QString &site : sites) {
-        const auto &cbResolv = [this, site, gw, mode, ips](const QHostInfo &hostInfo) {
+        const auto &cbResolv = [this, site, gw, mode, ips, remainingLookups, needFlush](const QHostInfo &hostInfo) {
             QStringList resolvedIps;
             for (const QHostAddress &addr : hostInfo.addresses()) {
                 if (addr.protocol() == QAbstractSocket::NetworkLayerProtocol::IPv4Protocol) {
@@ -245,7 +249,7 @@ void VpnConnection::addSitesRoutes(const QString &gw, amnezia::RouteMode mode)
             QStringList newIps;
             for (const QString &ip : resolvedIps) {
                 if (!ips.contains(ip)) {
-                    IpcClient::withInterface([&gw, &ip](QSharedPointer<IpcInterfaceReplica> iface) {
+                    IpcClient::withInterface([gw, ip](QSharedPointer<IpcInterfaceReplica> iface) {
                         iface->routeAddList(gw, QStringList() << ip);
                     });
                     newIps.append(ip);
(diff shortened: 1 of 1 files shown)


romanio427:
### Rebased onto current `dev` (post-#2811)

`dev` received #2811 in the same `VpnConnection::addSitesRoutes()` path, which caused the merge conflict.

**Resolution in `f6d5e26`:**
- kept multi-IP resolution + `addVpnSite(mode, site, newIps)` from #2811
- kept the stack-overflow fix: no `waitForFinished()` inside `QHostInfo` callbacks
- flush DNS **once**, asynchronously via `QRemoteObjectPendingCallWatcher`, and only if any new routes were installed (`needFlush`)
- capture `[gw, ip]` by value for `routeAddList` (replaces the previous `[&gw, &ip]` capture)

PR should be conflict-free against current `dev`. Runtime check on a large site list is still recommended.

vkamn:
thanks!

PR #2980, opened 2026-08-12
fix(i18n): restore the Russian translations dropped from ru_RU.ts

WkdXeqtr:
ru_RU.ts on dev has 237 entries in 28 contexts, all of them from C++, and the QML contexts are missing, so the pages themselves are untranslated. On the same commit uk_UA has 979 entries and zh_CN has 1104.

Looks accidental. lupdate over the C++ folders alone finds 216 texts in 28 contexts, which is what dev has, and adding client/ui/qml finds 1224 in 128. The QML files are not sources of the target, they only reach the build through qml.qrc.

Regenerated with lupdate over both, starting from the last revision that still had the QML contexts, and the gaps filled from my #2923. Nothing that is on dev today is removed or reworded, I checked entry by entry. lrelease gives 234 translations on dev and 1071 finished here.

The screenshots run the same strings, copied from the real pages, through the two compiled .qm files. Left is before, right is dev today.

| Before | Today on dev |
| --- | --- |
| translated | untranslated |

### DIFF
--- a/client/translations/amneziavpn_ru_RU.ts
+++ b/client/translations/amneziavpn_ru_RU.ts
@@ -2,49 +2,88 @@
 <!DOCTYPE TS>
 <TS version="2.1" language="ru_RU">
 <context>
+    <name>AllowedDnsController</name>
+    <message>
+        <source>The address does not look like a valid IP address</source>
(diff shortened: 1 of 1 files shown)


vkamn:
fixed here https://github.com/amnezia-vpn/amnezia-client/pull/3002

PR #2981, opened 2026-08-12
fix: windows connection stability

NickVs2015:
- Detect real sleep/resume via `PBT_APMSUSPEND`+`PBT_APMRESUMEAUTOMATIC` on Windows** (`windowsnetworkwatcher.cpp/.h`)
  `PBT_APMRESUMESUSPEND` is deprecated since Windows 8 and isn't reliably sent under Modern Standby. Switched to tracking a `PBT_APMSUSPEND` → `PBT_APMRESUMEAUTOMATIC` pairing before emitting `wakeup()`, so the signal fires on every real resume instead of only user-input-triggered ones, and doesn't fire spuriously. This alone cut reconnect-trigger bursts after sleep from ~6 down to ~1-2 per wake event.

- Make `WireguardProtocol::stop()` idempotent** (`wireGuardProtocol.cpp/.h`)
  `VpnConnection::disconnectFromVpn()` called `m_vpnProtocol->stop()` explicitly and then reset the `QSharedPointer`, which immediately ran the destructor — and the destructor unconditionally called `stop()` again, causing `deactivate()` to run twice per disconnect. Added an `m_stopped` guard so a second `stop()` call is a no-op.

- Correct leak-detector dtor macro and honor service poll timeout on Windows** (`windowstunnelservice.cpp`, `windowsservicemanager.cpp`)
  `WindowsTunnelService`'s destructor called `MZ_COUNT_CTOR` instead of `MZ_COUNT_DTOR` (copy-paste typo), so the leak tracker's bookkeeping for that pointer never got cleared correctly — eventually tripping `Q_ASSERT("leak.contains(ptr)")` and aborting the debug build. Also made `WindowsServiceManager::startService()`/`stopService()` return the actual result of polling for the target service state instead of just whether the initial `StartService`/`ControlService` call was accepted, so a timed-out wait is reported as failure instead of a false success.

- Correct `ControlService` call and ANSI error message decoding on Windows** (`windowsservicemanager.cpp`, `windowsutils.cpp`)
  `ControlService(..., NULL)` is invalid per the Win32 API contract (the third parameter must point to a real `SERVICE_STATUS`) and deterministically failed with `ERROR_INVALID_ADDRESS`. Also fixed `WindowsUtils::getErrorMessage()`, which decoded `FormatMessageA`'s ANSI-codepage output as UTF-8, turning Russian-locale error text into mojibake; now uses `QString::fromLocal8Bit` and includes the numeric error code in the log line.

- Fix `QRingBuffer` corruption under concurrent logging** (`common/logger/logger.cpp/.h`)
  The shared `Logger` writes to a single process-wide static `QFile`/`QTextStream` with no synchronization. `tunnel.dll` spins up dozens of WireGuard worker threads that log concurrently with the main thread, and unsynchronized writes to the same `QFile` from multiple threads corrupted Qt's internal write buffer, tripping `Q_ASSERT("bytes <= bufferSize")` in `qringbuffer.cpp` and crashing the service — most reproducible on the very first connect (cold tunnel bring-up = heaviest concurrent logging). Added a `QMutex` guarding every access to the shared file/stream (`LogStreamer` destructor, `init`, `deInit`, `clearLogs`, `getLogFile`/`getServiceLogFile`).

### DIFF
--- a/client/core/protocols/wireGuardProtocol.cpp
+++ b/client/core/protocols/wireGuardProtocol.cpp
@@ -50,6 +50,13 @@ WireguardProtocol::~WireguardProtocol()
 
 void WireguardProtocol::stop()
 {
+    // stop() is called both explicitly by callers and unconditionally from
+    // the destructor; guard against running deactivate() twice for one
+    // connection.
+    if (m_stopped) {
+        return;
+    }
+    m_stopped = true;
     stopMzImpl();
     return;
 }
@@ -75,5 +82,6 @@ ErrorCode WireguardProtocol::stopMzImpl()
 
 ErrorCode WireguardProtocol::start()
 {
+    m_stopped = false;
     return startMzImpl();
 }
--- a/client/core/protocols/wireGuardProtocol.cpp
+++ b/client/core/protocols/wireGuardProtocol.cpp
@@ -50,12 +50,10 @@ WireguardProtocol::~WireguardProtocol()
 
 void WireguardProtocol::stop()
 {
-    if (m_stopped) {
+    if (guardStop()) {
         return;
     }
-    m_stopped = true;
     stopMzImpl();
-    return;
 }
 
 ErrorCode WireguardProtocol::startMzImpl()
@@ -79,6 +77,6 @@ ErrorCode WireguardProtocol::stopMzImpl()
 
 ErrorCode WireguardProtocol::start()
 {
-    m_stopped = false;
+    resetStopGuard();
     return startMzImpl();
 }
--- a/client/core/protocols/wireGuardProtocol.cpp
+++ b/client/core/protocols/wireGuardProtocol.cpp
@@ -50,9 +50,6 @@ WireguardProtocol::~WireguardProtocol()
 
 void WireguardProtocol::stop()
 {
-    // stop() is called both explicitly by callers and unconditionally from
-    // the destructor; guard against running deactivate() twice for one
-    // connection.
     if (m_stopped) {
         return;
     }
--- a/client/platforms/windows/windowsutils.cpp
+++ b/client/platforms/windows/windowsutils.cpp
@@ -32,10 +32,6 @@ QString WindowsUtils::getErrorMessage(quint32 code) {
   return result;
 }
 
-QString WindowsUtils::getErrorMessage() {
-  return getErrorMessage(GetLastError());
-}
-
 // A simple function to log windows error messages.
 void WindowsUtils::windowsLog(const QString& msg) {
   quint32 code = GetLastError();
--- a/client/platforms/windows/windowsutils.cpp
+++ b/client/platforms/windows/windowsutils.cpp
@@ -27,8 +27,7 @@ QString WindowsUtils::getErrorMessage(quint32 code) {
       nullptr, code, MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT),
       (LPSTR)&messageBuffer, 0, nullptr);
 
-  std::string message(messageBuffer, size);
-  QString result(message.c_str());
+  QString result = QString::fromLocal8Bit(messageBuffer, static_cast<int>(size)).trimmed();
   LocalFree(messageBuffer);
   return result;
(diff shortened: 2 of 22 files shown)


client/core/protocols/wireGuardProtocol.cpp
```diff
@@ -50,6 +50,10 @@ WireguardProtocol::~WireguardProtocol()
 
 void WireguardProtocol::stop()
 {
+    if (m_stopped) {
```
ygurov: No other protocol have this mechanism. I would rather do it in base class, or remove it out of WireguardProtocol, and address it somewhere else

client/platforms/windows/windowsutils.cpp
```diff
@@ -39,8 +38,9 @@ QString WindowsUtils::getErrorMessage() {
 
 // A simple function to log windows error messages.
 void WindowsUtils::windowsLog(const QString& msg) {
-  QString errmsg = getErrorMessage();
```
ygurov, line 42: Remove unused `getErrorMessage()`

client/ui/qml/main2.qml
```diff
@@ -16,6 +16,31 @@ Window  {
...
+            busyIndicator.visible = false
+            PageController.disableControls(false)
+        }
+    }
+
+    Connections {
+        target: ConnectionController
+        function onSystemWoke() {
+            console.log("ConnectionController.systemWoke received")
+            root.clearStuckUiBlockIfNeeded("system wakeup")
```
ygurov: This hack does not fix anything.
Busy indicator/disable constrols is supposed to cover a slow operation, and the original root cause was not fixed.

common/logger/logger.cpp
```diff
@@ -68,22 +69,30 @@ bool Logger::init(bool isServiceLogger)
...
+            m_textStream.setDevice(&m_file);
+        }
+    }
+
+    if (!opened) {
         qWarning() << "Cannot open log file:" << logFileName;
         return false;
     }
 
-    m_file.setTextModeEnabled(true);
```
ygurov, line 77: what's the point? less readable IMO

common/logger/logger.cpp
```diff
@@ -68,22 +69,30 @@ bool Logger::init(bool isServiceLogger)
         return false;
     }
 
-    m_file.setFileName(appDir.filePath(logFileName));
-    if (!m_file.open(QIODevice::Append)) {
+    bool opened = false;
+    {
+        QMutexLocker locker(&m_fileMutex);
```
ygurov: If the logger could be called from simultaneous contexts, we would better mutex the whole calls. Otherwise, nothing is actually being blocked, and we still in the situation with simultaneous access

ygurov (changes requested):

client/core/protocols/wireGuardProtocol.cpp
z-elements: moved to VpnProtocol

client/platforms/windows/windowsutils.cpp
z-elements: removed

common/logger/logger.cpp
z-elements: refactoring this block

client/ui/qml/main2.qml
NickVs2015: removed

common/logger/logger.cpp
NickVs2015: fixed it

client/platforms/windows/windowsutils.cpp
NickVs2015: removed

common/logger/logger.cpp
NickVs2015: refactor this block

client/core/protocols/wireGuardProtocol.cpp
NickVs2015: moved to VpnConnection base class

PR #3019, opened 2026-08-22
Add ARM64 support for the Linux client

Broklam:
## Summary

Adds native ARM64 (`aarch64`) support for the Linux desktop client: the build scripts now work on ARM hosts, packages are named after the target architecture, and CI builds and uploads an ARM64 client.

Everything here was validated by building the client natively on an `aarch64` board (Raspberry Pi, Debian) — Conan builds every dependency from source for `armv8`, so no prebuilt ARM binaries are required.

## Changes

**`deploy/build.sh` — derive the Qt toolchain directory from the host architecture**

Qt names its Linux toolchain folder after the host arch, so an ARM64 install lands in `gcc_arm64`, not `gcc_64`. The script hardcoded `gcc_64` in both `QT_HOST_PATH` and `CMAKE_PREFIX_PATH`, so it could never find Qt on ARM. It now picks the directory from `uname -m`; behaviour on x86_64 is unchanged.

**`cmake/CPack.cmake` — name Linux packages after the target architecture**

`CPACK_PACKAGE_FILE_NAME` was hardcoded to `linux_x64`, which would have produced an ARM64 package labelled x64. It now uses `arm64` when `CMAKE_SYSTEM_PROCESSOR` is `aarch64`/`arm64`, and `x64` otherwise.

**`README.md` — document building on Linux ARM64**

Qt does not offer ARM64 desktop packages through the online installer, but they are published and `aqtinstall` can fetch them, so the exact command is written out. Also notes the memory-constrained case (limiting Conan and build parallelism to avoid the compiler being OOM-killed), and adds the missing `qtshadertools` entry to the module list — it is required at runtime by `Qt5Compat.GraphicalEffects`.

**`.github/workflows/deploy.yml` — build the Linux client for ARM64**

New `Build-Linux-Ubuntu-Arm64` job on the `ubuntu-22.04-arm` runner, mirroring the existing x64 job with three differences:

- `install-qt-action` does not expose the `linux_arm64` host, so `aqt` is driven directly.
- There are no ARM64 prebuilts on the `amnezia` Conan remote, so every dependency is built from source. `~/.conan2` is cached to keep subsequent runs cheap.
- No installer is produced — Qt Installer Framework has no ARM64 build — so the job uploads the `AmneziaVPN` client and `AmneziaVPN-service` binaries directly.

## Test plan

- Built the client natively on `aarch64` (Debian, Qt 6.10.1 via `aqtinstall`) — Conan resolved and built all dependencies from source, and both `AmneziaVPN` and `AmneziaVPN-service` link and run.
- x86_64 paths are untouched: `uname -m` falls through to `gcc_64` and CPack falls through to `linux_x64`.
- The new CI job has not run on this repo's runners yet; it needs a maintainer to approve the workflow run on the PR.


WkdXeqtr:
I have a Raspberry Pi and there is no aarch64 client, so I ended up setting up AmneziaWG by hand instead of using the app.

Headless is covered too - #2517 is already labelled in review. So between this PR and that one, the Linux side is effectively ready in either shape, desktop or headless. Can maintainers please look into this one too?

Issue #3037, opened 2026-08-25
Linux: AmneziaWG autoconnect fails after reboot due to missing exclusion route to VPN endpoint


highlander-ekb:
**Describe the bug**

On Kubuntu 24.04 with AmneziaVPN 5.0.1.5 and AmneziaWG, automatic VPN connection after system boot may fail.

Amnezia creates the `amn0` interface and installs the default split routes:

0.0.0.0/1 dev amn0 metric 1
128.0.0.0/1 dev amn0 metric 1

However, during the failed autoconnect attempt, the exclusion route to the VPN server endpoint is not created.

The log contains:

getGatewayAndIface: no gateway found
No default gateway available, skipping exclusion route

After that, Amnezia tries to establish the AmneziaWG connection, but the handshake cannot be completed and the client remains in the "Connecting" state indefinitely.

The log also contains:

Failed to send handshake initiation: network is unreachable

At this point there is no Internet connectivity.

If I manually click Disconnect and then Connect in the AmneziaVPN GUI, the connection succeeds immediately.

During the successful manual connection Amnezia correctly detects the physical gateway and creates the exclusion route:

Adding exclusion route for <VPN_SERVER_IP>/32
Gateway 192.168.1.1 for interface wlp0s20f3

After that the VPN and Internet connection work normally.

**To Reproduce**

1. Configure AmneziaVPN to automatically connect using AmneziaWG.
2. Reboot Kubuntu.
3. Log in and wait for AmneziaVPN autoconnect.
4. AmneziaVPN remains in the connecting state and Internet access is unavailable.
5. Check the routing table. The following routes exist:

0.0.0.0/1 dev amn0 metric 1
128.0.0.0/1 dev amn0 metric 1

but there is no host route for the VPN endpoint via the physical Wi-Fi gateway.
6. In AmneziaVPN manually click Disconnect.
7. Click Connect.
8. The VPN connects successfully and Internet access is restored.
9. The routing table now contains an exclusion route similar to:

<VPN_SERVER_IP> via 192.168.1.1 dev wlp0s20f3

**Expected behavior**

AmneziaVPN should not install routes that direct the VPN server endpoint itself into `amn0`.

Before redirecting traffic through `amn0`, the client should ensure that an exclusion route to the VPN endpoint through the physical network gateway exists.

If the default gateway is not yet available during autoconnect, AmneziaVPN should wait/retry instead of continuing with a configuration in which the VPN endpoint becomes unreachable.

[...]

Issue #2997, opened 2026-08-17
Linux: endpoint exclusion route skipped after switching servers due to failed default-gateway detection


samdark:
Related to #2048, but not fixed by #2840.

I have two AWG2 server connections in the Linux client:

- The first server works normally.
- The second server reports `Connected` and completes the handshake, but no traffic passes.
- The same second connection works on a phone connected to the same LAN.
- Recreating the second server from scratch did not help.

The client log shows that it fails to install the endpoint exclusion route:

```
Amnezia LinuxRouteMonitor : Adding exclusion route for <VPN_SERVER_IP>/32
Amnezia LinuxRouteMonitor : No default gateway available, skipping exclusion route
```

It then installs the full-tunnel routes:

```
Adding route to 0.0.0.0/1
Adding route to 128.0.0.0/1
```

Consequently, traffic to the VPN endpoint itself is routed into `amn0`. The initial handshake succeeds before the full-tunnel routes are installed, so the UI misleadingly reports that the connection is active.

The working server already had a correct endpoint route through the physical interface. Adding the missing route manually fixed the second connection immediately:

```
sudo ip route replace <VPN_SERVER_IP>/32 via <LAN_GATEWAY> dev <PHYSICAL_INTERFACE>
```

After this command, AWG2 traffic through the second server works normally.

The logs also contain:

```
Error: argument "amnvpnrt" is wrong: invalid table ID
mkdir: No such file or directory: /sys/fs/cgroup/net_cls/amnvpnexclusions/
```

Environment

- AmneziaVPN: `5.0.0.5` (`4d28f81`, built July 26, 2026)
- OS: Ubuntu 26.04 LTS (Resolute Raccoon)
- Kernel: `7.0.0-29-generic`
- Architecture: x86_64
- VPN interface: `amn0`
- Protocol: AmneziaWG v2

This appears to be a Linux client bug in default-gateway detection or endpoint exclusion-route management, particularly when multiple server connections are configured or switched between.

#2840 does not cover the missing endpoint exclusion-route variant.

cucumberian:
Похоже я тоже словил этот баг в новой версии, вот репорт от квена:

Related to #2997 and #2048 — I was able to pin down the exact trigger for the "No default gateway available" failure on my machine.

## Environment

- AmneziaVPN: 5.0.1.5 (linux x64, .run installer)
- OS: Linux Mint 22.3 (Zena), kernel 6.8.0-138-generic, x86_64
[...]

Issue #2048, opened 2025-12-09
AmneziaWG Protocol Fails to Route Traffic on Linux Despite Successful Connection


anothervoyager:
**Describe the bug**  
AmneziaWG protocol fails to provide internet access on Linux distributions despite the AmneziaVPN client reporting a successful connection. Browsers cannot load any websites, while the same configuration works correctly on Windows 11. Other protocols (e.g., Xray) function properly on the same Linux systems.

**To Reproduce**  
1. Install AmneziaVPN version 4.8.11.4 on a Linux system (tested on Arch Linux, Ubuntu 24.04, Debian 13).  
2. Configure a server using the AmneziaWG protocol.  
3. Connect to the server via the AmneziaVPN client.  
4. Observe that the client displays "Connected", but no websites load in the browser.  

**Expected behavior**  
After a successful connection using the AmneziaWG protocol, internet traffic should be properly routed through the VPN, and websites should load normally—consistent with behavior on Windows 11.

**Log files**  
Relevant log snippet (from Arch Linux):  
```
[2025-12-09 12:28:24.089Z] [DEBUG] Amnezia  : IpcClient::init succeed  
[2025-12-09 12:28:24.094Z] [INFO] Amnezia  : Logging has enabled on AmneziaVPN version 4.8.11.4 3b3a0aa  
[2025-12-09 12:28:24.094Z] [INFO] Amnezia  : Arch Linux (x86_64)  
[2025-12-09 12:28:28.968Z] [DEBUG] Amnezia  : VpnConfigurator::getDnsForConfig "1.1.1.1" "1.0.0.1"  
[2025-12-09 12:28:28.969Z] [DEBUG] Amnezia  : "ConnectToVpn, Server index is 0, container is amnezia-awg, route mode is" Settings::VpnOnlyForwardSites  
[2025-12-09 12:28:28.970Z] [DEBUG] Amnezia LocalSocketController : Daemon connected  
[2025-12-09 12:28:30.515Z] [DEBUG] Amnezia LocalSocketController : Parse command: connected  
[2025-12-09 12:28:30.515Z] [DEBUG] Amnezia LocalSocketController : Handshake completed with: B6GAa2Z/0+91ap3orgi0Z2yPL3hTAakytAkDEyhemTo=
```  

**Screenshots**  
Not included, but UI shows "Connected" status while browser requests time out.

**Desktop (please complete the following information):**  
[...]

Issue #2549, opened 2026-04-30
route to VPS not added when using hosted solution and PC has secondary route tables


rgalexey:
I was playing with ip_route2 while using Amnezia Client  with VPS.

So noticed following thing.
For example IP of my VPS is 5.6.7.8
So when Amnezia connected to VPN it always add route like
ip r add 5.6.7.8 via MY_ROUTER_IP dev eth0

But if I will add some route to other routing table
For example 
ip r add 1.1.1.1 via MY_OTHER_ROUER dev eth0 table myroutetable

Amnezia in that case will fail to add route 
ip r add 5.6.7.8 via MY_ROUTER_IP dev eth0

So basically VPN wil not work.
I had add in such case that route manualy.

Somehow Amnezia confused when any other route tables exist.
And shows error 'Amnezia LinuxRouteMonitor : No default gateway available, skipping exclusion route'

OS: Fedora 44
Amnezia Client: AmneziaVPN_4.8.15.4

Here is log while fail is happaning:
Apr 30 22:56:27 phantom AmneziaVPN-service.sh[281087]: [2026-04-30 19:56:27.519Z] [DEBUG] Amnezia LinuxRouteMonitor : Adding exclusion route for 5.6.7.8/32
Apr 30 22:56:27 phantom kernel: amn0: entered promiscuous mode
Apr 30 22:56:27 phantom AmneziaVPN-service.sh[281087]: [2026-04-30 19:56:27.520Z] [WARNING] Amnezia LinuxRouteMonitor : No default gateway available, skipping exclusion route
Apr 30 22:56:27 phantom AmneziaVPN-service.sh[281087]: [2026-04-30 19:56:27.520Z] [DEBUG] Amnezia WireguardUtilsLinux : Configuring peer ****************/NmFckHlkvDcowQ= via 5.6.7.8
Apr 30 22:56:27 phantom AmneziaVPN-service.sh[281087]: [2026-04-30 19:56:27.520Z] [DEBUG] Amnezia LinuxRouteMonitor : Adding exclusion route for 5.6.7.8/32
Apr 30 22:56:27 phantom AmneziaVPN-service.sh[281087]: [2026-04-30 19:56:27.520Z] [WARNING] Amnezia LinuxRouteMonitor : No default gateway available, skipping exclusion route
[...]

PR #3052, opened 2026-08-25
fix: read the full netlink route dump when detecting the default gateway

izhddm:
## Problem

On Linux, site-based split tunnelling silently stops working on hosts with many network interfaces. The exclusion list is still shown in the UI, but every route is skipped and all traffic goes through the tunnel. The only trace is in the service log:

```
LinuxRouteMonitor : Adding exclusion route for 84.252.149.0/24
LinuxRouteMonitor : No default gateway available, skipping exclusion route
Error in received packet: Numerical argument out of domain
```

Because nothing surfaces in the UI, this reads as a misconfiguration rather than a failure.

## Related reports

The failing log line appears in several open issues, where the consequence goes beyond split tunnelling: when the exclusion route to the VPN endpoint itself cannot be installed, endpoint traffic is routed into the tunnel, so the connection either passes no traffic or never completes a handshake.

Worth separating two different ways `getGatewayAndIface()` can come back empty, because this PR only addresses the first:

1. **A default route exists but is not found** - the dump overflows the buffer. That is the case measured here and what this patch fixes.
2. **No default route exists yet** - e.g. autoconnect racing network setup at boot. No parser can find what is not in the table; that needs a retry or a wait, not this change.

- #3090 - Ubuntu 24.04, 5.0.1.5 vs 4.8.21 on the same machine: `Error in received packet` and `No default gateway available` ten times each on 5.0.1.5, none on 4.8.21; endpoint routed into the tunnel, `sendmmsg: message too long`. Fits case 1; the route count on that host is not in the report yet.
- #2997 - second server reports `Connected` but passes no traffic, with `No default gateway available, skipping exclusion route` on a running system. Fits case 1.
- #3037 - autoconnect after reboot fails; reconnecting by hand then succeeds immediately. That points at case 2 rather than at this bug, since the route count does not change between the two attempts. Listing it because the log line is identical and the two causes are easy to confuse.
- #2048 - same family of symptoms; I could not tell which case applies from the report.

Verified only on my own host, so I am not marking any of these as fixed - maintainers can confirm which reports the change actually covers.

## Root cause

`NetworkUtilities::getGatewayAndIface()` accumulates the whole `RTM_GETROUTE` dump into a single 8 KiB buffer:

```c
received_bytes = recv(sock, ptr, sizeof(buffer) - msg_len, 0);
```

Two things make the dump large:

- `rtm_family` is never set (the request buffer is zeroed), so the kernel also returns IPv6 routes;
- container bridges, virtual machines and VPN interfaces each contribute routes.

Once `msg_len` reaches `sizeof(buffer)`, `recv()` is called with a zero-sized remainder, `NLMSG_OK()` fails and the function returns an empty pair. `LinuxRouteMonitor::rtmSendRoute()` then skips every `RTN_THROW` exclusion route.

Measured on an affected host (22 docker bridges, libvirt, tailscale, an active tunnel):

| | routes | dump size |
|---|---|---|
| Current request (`AF_UNSPEC`) | 118 | 10292 B |
| Buffer | | 8192 B |

The threshold is roughly 94 routes at ~87 bytes per entry. Below it everything works, which is why this does not show up on a typical desktop.

## Fix

- Parse each netlink datagram as it arrives instead of accumulating the whole dump, removing the dependency on any fixed dump size.
- Set `rtm_family = AF_INET`. The gateway is decoded with `inet_ntop(AF_INET, ...)`, so an IPv6 route reaching that code would have produced a bogus IPv4 address.
- Require `rtm_dst_len == 0` when selecting the route. The previous code accepted the first route carrying a gateway, which is not necessarily the default one. This matters for a VPN client in particular: `0.0.0.0/1` and `128.0.0.0/1` are installed precisely to shadow the default route.
- Prefer the lowest metric among several default routes, matching kernel behaviour.

Behaviour is unchanged on hosts where the dump already fitted into the buffer. Logging is left as it was, to keep the change focused.

## Relation to earlier work

This is not the first report of the problem, and the fix deliberately differs in approach.

- **#2862** (merged, shipped in 5.0.0.5) raised `BUFFER_SIZE` from 100 to 8192 and fixed the receive loop, which until then tested `NLM_F_MULTI` on the *request* header rather than the response and therefore exited after the first `recv()`. In 4.8.x the function effectively parsed only the first ~100 bytes of the dump - a single route message. That happened to be the default route whenever it comes first in the main table, which is the case at the moment the endpoint exclusion route is installed, before the tunnel's `0.0.0.0/1` routes exist. So 4.8.21 found the gateway by accident of dump ordering; #2862 made the loop read the whole dump and thereby exposed the overflow on hosts whose dump exceeds 8 KiB. This is why users report 4.8.21 working on a machine where 5.0.x does not (#3090).
- **#2287** (open since February, no review yet) diagnosed the same overflow back when the buffer was 100 bytes, and proposed 16384 plus an `AF_INET` filter. It also carries fixes outside the scope of this PR (socket fd leaks, `nlmsg_seq` validation, extra logging in the daemon) that remain worth taking on their own.

Both keep a fixed buffer, so the dump size stays a hard limit and the threshold merely moves. The host in this report already exceeds 8192 with a routine developer setup, and adding interfaces would eventually exceed 16384 too. Parsing datagrams as they arrive removes the limit rather than raising it, which is why this PR takes that route instead of picking a larger constant.

If maintainers prefer to land #2287 first, this change rebases onto it without conflict beyond the same function body.

## Testing

The patched block was extracted, compiled standalone with `-Wall -Wextra -Wpedantic` (no warnings) and run on the affected host:

```
candidate: 192.168.0.1 dev wlp9s0 metric 600
datagrams received: 2, candidates: 1
RESULT: gateway 192.168.0.1 via wlp9s0 (metric 600)
```

The same host returned no gateway before the change. With the gateway resolved, the exclusion routes are installed and site-based split tunnelling works as configured.

### DIFF
--- a/client/core/utils/networkUtilities.cpp
+++ b/client/core/utils/networkUtilities.cpp
@@ -23,6 +23,8 @@
     #include <sys/ioctl.h>
     #include <sys/socket.h>
     #include <unistd.h>
+    #include <climits>
+    #include <cstring>
 #endif
 #if defined(Q_OS_MAC) && !defined(Q_OS_IOS) && !defined(MACOS_NE)
     #include <sys/param.h>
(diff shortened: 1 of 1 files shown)


izhddm:
A status update rather than a ping, since several things have been established since this was opened.

* #3090 was reproduced by the reporter with numbers: 60 IPv4 and 170 IPv6 routes, which by the per-message sizes below is roughly 23 KiB of dump against an 8 KiB buffer.
* #2549 turns out to be the same function failing for a different reason on 4.8.15, where the receive buffer was 100 bytes and the multipart flag was tested on the request header. One route in a foreign table was enough to displace the main-table default route out of that window, because the kernel emits tables in its own order rather than by id — on my kernel a custom table comes out before main. #2862 fixed that variant and introduced the ceiling this PR removes.
* #3125 was diagnosed with that reproducer today and turned out to be the same overflow: 16 IPv4 and 68 IPv6 routes, 8192 of 8192 bytes accumulated. Removing a couple of Docker stacks brought the connection back, which is about as direct a confirmation as this bug allows. The reporter's second machine sits at 8072 of 8192 — 120 bytes of headroom against a 116-byte route message, so one more container breaks it too.
* #2997 contains an independent analysis that lands on the same lines of code.
* Calibration, measured here: a route message is 60 bytes for IPv4 and 116 for IPv6, so the 8 KiB ceiling falls at roughly 135 routes on an IPv4-only table and around 70 on an IPv6-heavy one. Hosts with containers cross it easily: every container veth adds exactly three IPv6 route messages (link-local prefix, local address, multicast) and no IPv4 ones, about 350 bytes each, so roughly 23 containers fill the buffer on their own. Every exclusion route that does get installed enlarges the next dump as well, since each `addExclusionRoute()` performs its own lookup.
* A no-build reproducer is available in https://github.com/amnezia-vpn/amnezia-client/issues/3125#issuecomment-5613110287. It replays the identical netlink exchange and reports which branch the current function takes, which also separates the three distinct ways it can return an empty pair.

The change itself stays inside `getGatewayAndIface()`: parse each datagram as it arrives, request `AF_INET` only, require an actual default route and prefer the lowest metric. It also removes the netlink descriptor leak on the error paths as a side effect.

@NickVs2015 you wrote the previous change to this function (#2862, netlink message lookup and buffer size) and @vkamn merged it — would either of you have time to take a look?

McSmog:
Confirmed this fixes the issue on my Linux machine with AmneziaVPN 5.0.3.0 and multiple Docker networks/routes.
The VPN server was reachable and the tunnel could handshake, but the GUI connection left me without internet access. The service logged `Error in received packet: Numerical argument out of domain` and `No default gateway available, skipping exclusion route`, even though `ip route` showed a valid default gateway. Disabling split tunneling did not help.
I built the 5.0.3.0 service with this PR applied and replaced only the local background service. Internet access through the AmneziaVPN GUI now works. Thank you for the fix — please merge and release it so other users don’t have to build the service themselves.

PR #3054, opened 2026-08-26
Fix: xray reconnect crash

yp-amnezia:


client/vpnConnection.h
```diff
@@ -92,6 +92,8 @@ protected slots:
 
    Vpn::ConnectionState m_connectionState;
 
+   bool m_transitionInFlight = false;
```
ygurov: The guard is supposed to protect from "simultaneous access", so no callback is left orphaned.
I think races/caching is highly-likely the case in here. So let's consider making it atomic

PR #3099, opened 2026-08-31
fix(windows): sanitize process ancestry before driver registration

Rain0Ash:
## Problem

The Windows client builds a process-parent map from a live Toolhelp snapshot and sends it to the split-tunnel driver by PID. Between process enumeration and registration, a PID can be recycled; an unreadable or stale entry can therefore point to an unrelated process, including itself or another node in a cycle.

The driver uses these relationships while applying inherited split state. A cyclic or stale parent link can make the kernel-side ancestry walk non-terminating, leaving DeviceIoControl(IOCTL_SET_CONFIGURATION) blocked and the AmneziaVPN service unresponsive. This is consistent with the intermittent WindowsSplitTunnel::excludeApps hangs and high CPU reported in amnezia-vpn/amnezia-client#1996.

This is a strongly supported root-cause hypothesis, not a claim that the original report included runtime proof of a cycle.

## Fix

- Zero-initialize PROCESSENTRY32W before Toolhelp calls.
- Use Process32FirstW and abort registration if the snapshot cannot be read.
- Keep a parent link only when the parent is present in the same snapshot and its creation time is strictly older than the child.
- Clear invalid parent PIDs before sending the process graph to the driver.

This prevents stale PID reuse and malformed ancestry from entering the kernel graph. The companion driver PR adds cycle detection as defense in depth and keeps cyclic processes in the safe non-split state.

The change is intentionally limited to process snapshot construction; it does not alter split-tunnel policy semantics for valid process trees.

## Validation

The patch was reviewed against the existing Toolhelp snapshot and IOCTL contract. Invalid snapshot reads and recycled-PID relationships are rejected before IOCTL_SET_CONFIGURATION is called.

Related issue: amnezia-vpn/amnezia-client#1996
Companion driver PR: https://github.com/mullvad/win-split-tunnel/pull/56

### DIFF
--- a/client/platforms/windows/daemon/windowssplittunnel.cpp
+++ b/client/platforms/windows/daemon/windowssplittunnel.cpp
@@ -149,6 +149,16 @@ ProcessInfo getProcessInfo(HANDLE process, const PROCESSENTRY32W& processMeta) {
   return pi;
 }
 
+bool retainParentRelationship(const ProcessInfo& parent,
+                              const ProcessInfo& child) {
+  const auto value = [](const FILETIME& time) {
+    return (static_cast<std::uint64_t>(time.dwHighDateTime) << 32) |
+           time.dwLowDateTime;
+  };
+  return value(parent.CreationTime) != 0 && value(child.CreationTime) != 0 &&
+         value(parent.CreationTime) < value(child.CreationTime);
+}
+
 QString normalizeExecutablePath(const QString& path) {
   QString normalized = path.trimmed();
   if (normalized.startsWith("file:", Qt::CaseInsensitive)) {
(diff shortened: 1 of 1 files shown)


Gerk17:
Tested together with #3112 on a real laptop: Windows 11, AmneziaWG, app-based split tunneling in "exclude selected apps" mode. Build: 5.0.1.5 + both commits of #3112 + this PR, built with the project's GitHub Actions workflow in a fork ([run](https://github.com/Gerk17/amnezia-client/actions/runs/34675869497)).

No regressions: excluded apps worked normally, including across network switches and sleep/wake (details in https://github.com/amnezia-vpn/amnezia-client/pull/3112#issuecomment-5644178574). I haven't hit the `excludeApps` hang from #1996 on this machine, so I can't confirm the fix itself. I can only confirm that it doesn't break split tunneling in this setup.

CI: Windows and Linux builds are green. iOS/macOS fail at the signing-certificate step because the fork has no Apple secrets, and Android never starts because `android-runner` is self-hosted. This PR only changes `client/platforms/windows/daemon/windowssplittunnel.cpp`.

PR #3112, opened 2026-09-01
fix(windows): refresh split tunnel addresses and support all address modes

Rain0Ash:
## Problem
Split tunnel addresses are registered only at start. On route / interface / address changes (Wi-Fi switch, VPN reconnect, address renewal) the registered addresses go stale and split tunneling breaks until reconnect. Additionally the client computes only a subset of the driver's splitting modes, so legitimate IPv4/IPv6 availability combinations misconfigure the driver.

## Fix
- Subscribe to NotifyRouteChange2 / NotifyIpInterfaceChange / NotifyUnicastIpAddressChange; debounce 250 ms; retry up to 3 times; wait for address stabilization before re-registering.
- Recompute availability (internet/tunnel x v4/v6) and select the matching driver splitting mode (1-9); pick the physical adapter from the server endpoint passed into start().
- Replace isRunning() with isOperational() for accurate state checks.

## Validation
- Network switch / adapter toggling / ipconfig renew during an active tunnel: split apps keep working.
- All 9 modes covered by compile-time static_asserts.
- Together with the companion driver PR: Java apps (ex. Minecraft) work in MODE_4.

## Related issue
mullvad/mullvadvpn-app#2972
## Companion driver PR
mullvad/win-split-tunnel#57 (MODE_4 IPv6 TCP bind fix)

### DIFF
--- a/client/platforms/windows/daemon/windowsdaemon.cpp
+++ b/client/platforms/windows/daemon/windowsdaemon.cpp
@@ -22,7 +22,6 @@
 #include "logger.h"
 #include "platforms/windows/daemon/windowsfirewall.h"
 #include "platforms/windows/daemon/windowssplittunnel.h"
-#include "platforms/windows/daemon/windowssplittunnelcontracts.h"
 #include "windowsfirewall.h"
 
 #include "core/utils/networkUtilities.h"
@@ -54,9 +53,15 @@ WindowsDaemon::~WindowsDaemon() {
 void WindowsDaemon::prepareActivation(const InterfaceConfig& config, int inetAdapterIndex) {
   // Before creating the interface we need to check which adapter
   // routes to the server endpoint
+  m_serverEndpoint = QHostAddress(config.m_serverIpv4AddrIn);
+  if (m_serverEndpoint.isNull()) {
+    m_serverEndpoint = QHostAddress(config.m_serverIpv6AddrIn);
+  }
   if (inetAdapterIndex == 0) {
-      auto serveraddr = QHostAddress(config.m_serverIpv4AddrIn);
-      m_inetAdapterIndex = NetworkUtilities::AdapterIndexTo(serveraddr);
+      m_inetAdapterIndex =
+          m_serverEndpoint.protocol() == QAbstractSocket::IPv4Protocol
+              ? NetworkUtilities::AdapterIndexTo(m_serverEndpoint)
+              : -1;
   } else {
       m_inetAdapterIndex = inetAdapterIndex;
   }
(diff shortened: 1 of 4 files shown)


Gerk17:
Confirming this bug on a real setup. Hitting it every time the laptop changes networks.

**Environment:** Windows 11, AmneziaVPN 5.0.1.5, AmneziaWG, app-based split tunneling in "exclude selected apps" mode (browsers, RustDesk, Synology Drive, etc.). Kill switch disabled.

**Repro:** connect on Wi-Fi network A, put the laptop to sleep, wake it up on Wi-Fi network B (different subnet, new DHCP address). The client shows "Connected" and reconnects on its own.

**Result:**
- Tunneled apps work fine.
- Excluded apps have no connectivity at all. The browser shows `ERR_NAME_NOT_RESOLVED`. Opening a literal IP (`http://1.1.1.1`) in an excluded Edge fails with `ERR_ADDRESS_INVALID` (WSAEADDRNOTAVAIL), so it's not just DNS: sockets of excluded processes get bound to an address that no longer exists on the host.
- `curl --interface <new Wi-Fi IP> http://1.1.1.1` from a non-excluded process works, so the physical path is fine.
- A manual disconnect/connect in the client fixes it immediately, until the next network change.

This matches the description here: IPs are registered with the split tunnel driver only in `WindowsSplitTunnel::start()`, and nothing refreshes them on route/address changes. Updating 4.8.14.5 → 5.0.1.5 didn't help. The relevant code is unchanged in current `dev` as far as I can tell.

It would be great to get this reviewed and included in the next release. For roaming laptops, app-based split tunneling is effectively broken without it.

Gerk17:
Tested this PR on a real laptop: Windows 11, AmneziaWG, app-based split tunneling (excluded: browsers, RustDesk, etc.). Build: 5.0.1.5 + both commits of this PR + #3099, built with the project's GitHub Actions workflow in a fork.

5.0.1.5 reliably broke here: after waking up in a different network, excluded apps failed with `ERR_ADDRESS_INVALID` until a manual reconnect. With this PR, excluded apps kept working without any manual reconnect in all of these cases:
- switching between home Wi-Fi and a phone hotspot (two different carriers);
- sleep → hotspot turned off → wake up on home Wi-Fi (auto-reconnect).

The service log matches the design:
- when the endpoint route falls into the tunnel during the switch, the config is zeroed (driver goes to READY);
- after address stabilization it is re-registered: mode 4 on IPv4-only Wi-Fi, mode 1 on the hotspot with IPv6;
- that includes a 4 → 1 transition without reconnect when IPv6 appeared a few seconds after connect.

Minor notes:
- commit 334e6905 doesn't build on its own: it includes the nonexistent `windowssplittunnelcontracts.h`, which the next commit removes. Worth squashing.
- the `Split-tunnel address availability` debug line is logged on every refresh and is quite noisy.

CI note: in the fork build ([run](https://github.com/Gerk17/amnezia-client/actions/runs/34675869497)) Windows and Linux are green. The iOS/macOS jobs fail at the signing-certificate step because the fork has no Apple secrets. The Android jobs never start because `android-runner` is a self-hosted runner that forks don't have. Neither is related to this PR, which only touches `client/platforms/windows/daemon/`.

PR #3114, opened 2026-09-01
fix(windows): eliminate route-change storms with large split-tunnel e…

hgminusone:
…xclusion lists

Importing a large IP exclusion list (~8.6k prefixes) made the Windows client unusable: connecting stalled for many minutes, disconnecting hung, and after a forced kill the client could not reconnect until reboot.

Three related defects in WindowsRouteMonitor:

1. Quadratic route installation. Every added exclusion route refetched the full routing table and interface metrics. Measured on stock: 500 prefixes = 1.4s, 1000 = 6.4s, 2000 = 33.4s (the kernel itself installs 500 routes in 146ms). Fix: batched add/delete entry points that snapshot the table once per address family; the teardown loop additionally refetched the table per deleted prefix.

2. Self-sustaining rescan storm. The monitor subscribes to NotifyRouteChange2, but its own rescan rewrites routes (captured clones, flapping next-hops), waking itself: with a large list the monitor rescanned ~4 times/second for the whole session, starving the IPC queue (deactivate took minutes to be processed). The callback filter cannot drop these notifications because the notification rows carry unreliable Protocol/Metric fields. Fix: unsubscribe for the duration of bulk operations and of the rescan itself, then run exactly one manual rescan; real network changes still trigger it via the restored subscription.

3. Orphaned exclusion routes. After a hard service kill the routes stayed in the system table pointing at the dead tunnel's next-hop, blackholing matching traffic (including the VPN server handshake) until reboot. Fix: purge NETMGMT+EXCLUSION_ROUTE_METRIC rows at monitor creation - that pair is only ever produced by this code.

Result with an 8658-prefix list: connect in 4-7s (route installation 0.85s), idle monitor after connect, teardown in seconds, reboot no longer needed after a crash.

### DIFF
--- a/client/daemon/daemon.cpp
+++ b/client/daemon/daemon.cpp
@@ -132,9 +132,7 @@ bool Daemon::activate(const InterfaceConfig& config) {
   }
 
   // Configure routing for excluded addresses.
-  for (const QString& i : config.m_excludedAddresses) {
-    addExclusionRoute(IPAddress(i));
-  }
+  addExclusionRouteList(config.m_excludedAddresses);
 
   // Add the peer to this interface.
   if (!wgutils()->updatePeer(config)) {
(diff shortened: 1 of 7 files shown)


yupic:
There is a similar PR: https://github.com/amnezia-vpn/amnezia-client/pull/3021

But the developers don’t seem very active when it comes to reviewing and merging pull requests...

PR #3151, opened 2026-09-10
Fix: Win xray app split tunneling

yp-amnezia:


client/core/protocols/xrayProtocol.cpp
```diff
@@ -295,6 +301,13 @@ ErrorCode XrayProtocol::setupRouting()
                 }
 
                 if (m_routeMode == amnezia::RouteMode::VpnAllSites) {
+#ifdef Q_OS_WIN
+                    auto routeAddDefault = iface->routeAddDefault(tunName);
```
ygurov: What do we need this one for? Why do we do this for Windows and on Xray provider only?

ygurov: Looks like a hack IMO

Issue #3091, opened 2026-08-29
[Android 16][XRay/VLESS WS+TLS] IPv6 endpoint causes TLS handshake failure with no IPv4 fallback; config edits are discarded


MaratGaZa:
## Describe the bug

AmneziaVPN reports `Connected` for an imported VLESS WebSocket + TLS profile,
but no traffic passes.

The endpoint domain is proxied by Cloudflare and has both A and AAAA records.
AmneziaVPN selects an IPv6 address. The bundled Xray core repeatedly reports:

```text
transport/internet/websocket: failed to dial to [<CLOUDFLARE_IPV6>]:443
> remote error: tls: handshake failure
```

The connection does not reach the server-side VLESS/Xray service. The same
original VLESS profile works in Hiddify and Happ on the same Android device and
network. It also works from Hiddify on macOS.

## Steps to reproduce

1. Import a VLESS URI whose endpoint is a Cloudflare-proxied domain with A and
   AAAA records.
2. Use `type=ws`, `security=tls`, a valid `sni`, WebSocket `host`, path, UUID and
   port 443.
3. Disable both application and site split tunneling.
4. Connect on Android 16.
5. Observe that the UI changes to `Connected`, but Internet access is absent.
6. Export the AmneziaVPN log.

## Actual behavior

- Android VPNService, TUN, tun2socks and Xray start.
- The UI reports `Connected` before the outbound TLS session succeeds.
- The endpoint is resolved to IPv6.
- Xray repeatedly receives `remote error: tls: handshake failure`.
- No IPv4 fallback occurs.
- The server-side Xray journal receives no matching connection.

## Confirmed workaround

A new VLESS URI was imported with a numeric Cloudflare IPv4 address in the
endpoint position. The original domain was retained in TLS SNI/serverName and
the WebSocket Host header. Traffic started working immediately.

Conceptual difference:

```text
Failing:  vless://<UUID>@<DOMAIN>:443?...&sni=<DOMAIN>&host=<DOMAIN>...
Working:  vless://<UUID>@<CLOUDFLARE_IPV4>:443?...&sni=<DOMAIN>&host=<DOMAIN>...
```

No server-side change was required.

## Configuration persistence problem

As a diagnostic attempt, the following was added to `streamSettings`:

```json
"sockopt": {
  "domainStrategy": "UseIPv4"
}
```

The editor initially displayed the modified JSON, but after the first
connection attempt the settings returned to their defaults. The runtime log
[...]

PR #3169, opened 2026-09-14
fix(android): preserve Xray TLS and HTTP host identit

HeisLuka:
## Summary

Fix Android Xray pre-resolution so that resolving the transport endpoint does
not rewrite TLS/HTTP identity fields elsewhere in the native Xray JSON.

Current Android code serializes the whole Xray config and then performs a
global hostname-to-IP string replacement. For VLESS WS + TLS profiles where
the same domain is used as the endpoint, TLS SNI and WebSocket Host, that can
turn all three values into the resolved IP and cause CDN TLS handshakes to fail.

This change keeps pre-resolution, but applies the resolved address only to
structured Xray endpoint fields:

- `outbounds[*].settings.vnext[*].address`
- `outbounds[*].settings.servers[*].address`

TLS `serverName`, REALITY `serverName`, WebSocket `Host`, XHTTP host values and
unrelated JSON strings are left untouched.

## Why this matches #3091

The workaround reported in #3091 uses a numeric IPv4 transport endpoint while
preserving the original domain in TLS SNI and WebSocket Host. That bypasses the
global replacement and restores traffic.

This patch preserves the same endpoint/identity separation automatically.

## Scope

This PR fixes only the Android hostname-to-IP rewrite part of #3091.

It intentionally does not address the separate issues reported there around:

- IPv4/IPv6 fallback
- the `Connected` health state
- raw native-config persistence
- IPv6 route exclusion prefixes

The reporter has offered to retest on the same Android 16 device once a build
is available.

Related to #3091


MaratGaZa:
Thanks @HeisLuka for following up on #3091 with this focused fix.

The change in `8f7f175` looks consistent with my reported workaround: update the transport address while preserving TLS SNI and WebSocket Host. I also appreciate keeping the other issues separate.

One suggestion before merging: could you add regression tests for `replaceResolvedEndpointAddresses` covering:

- The same hostname in the endpoint, SNI, and Host, with both IPv4 and IPv6 resolution results: only the endpoint address should change.
- Both `vnext` and `servers`, including multiple entries/outbounds and unrelated addresses.
- The hostname appearing in unrelated JSON strings or as a substring: those values should remain unchanged.
- Already-numeric endpoints and missing optional arrays.

Ideally, the main identity-preservation test would fail with the old global replacement and pass with this implementation.

I have not built or run this patch yet. Could you provide an Android test APK, or point me to a build artifact, with its source commit identified? I can retest the original domain-based VLESS WS+TLS profile on the same Redmi running Android 16, over both Wi-Fi and mobile data, and report actual traffic results with sanitized logs.

That would validate this fix separately from the remaining address-fallback, IPv6 route-exclusion, raw-config persistence, and connection-health issues.

PR #3170, opened 2026-09-14
fix: handle large netlink route dumps in getGatewayAndIface on Linux

Ramurio:
## Problem

On Linux, `NetworkUtilities::getGatewayAndIface()` fails to read the kernel routing table whenever the netlink route dump exceeds the preallocated 8 KiB buffer — e.g. on a machine running Docker (bridge networks, veth links with IPv6 addresses). With ~20 containers the dump measures ~9.6 KiB.

The consequences cascade:

1. `recv()` silently truncates datagrams that don't fit the remaining capacity;
2. once the buffer is full, `recv(sock, ptr, 0, ...)` is called, which returns **0** — the code treats this as a protocol error and bails out;
3. `getGatewayAndIface()` returns an empty gateway;
4. `LinuxRouteMonitor` logs `No default gateway available, skipping exclusion route` and never installs the endpoint exclusion route;
5. WireGuard/AmneziaWG traffic to the VPN server gets routed **into the tunnel itself** and dies with `sendmmsg: message too long` (tunnel MTU < packet size);
6. the client shows *connected* but no traffic flows — on every server, every network, any client version.

Additional issues fixed along the way:

- the receive loop condition compared the reply against `nlmsg_pid = getpid()` from the request header; the kernel assigns the real port id to the (unbound) socket, and that pid may already be taken by another netlink socket in the same process (e.g. the one owned by `LinuxRouteMonitor`), so the loop could never terminate correctly;
- `perror("Error in received packet")` printed a stale, unrelated errno (e.g. `Numerical argument out of domain`), misleading debugging; now the actual netlink error is logged for `NLMSG_ERROR`;
- the netlink socket leaked on early returns.

## Evidence

Reproduced standalone while restarting 21 Docker containers:

- old code: first `recv()` fails once the dump exceeds the buffer (empty datagram returned by `recv(sock, ptr, 0, ...)`);
- dump arrives as 3 datagrams: `1036 + 8584 + 20 (NLMSG_DONE)` bytes — the second one alone is larger than the old buffer;
- patched code: 75/75 successful gateway lookups during a full container-restart storm, and the VPN connects with Docker running (previously: `No default gateway available` → route loop → dead tunnel).

## Changes

- dump buffer: 8 KiB stack → 128 KiB heap, with an explicit 'dump too large' bailout;
- never call `recv()` with zero capacity; treat an empty datagram as end of transfer;
- `nlmsg_pid = 0` in the request (let the kernel fill the real port id); rely on `NLMSG_DONE` / `NLM_F_MULTI` to end the loop;
- log the real netlink error instead of stale errno;
- close the socket on all early returns.


izhddm:
Hi, this is the same bug as #3052, which I opened in August. Just letting you know so the maintainers don't review both separately. Your write-up matches what I found. The measurements with 21 containers are useful, and so is the 75/75 run during a restart storm.

I went a different way. #3052 doesn't use a larger buffer; it parses each datagram as it arrives, so there is no upper limit on the dump size at all. It also asks the kernel for `AF_INET` only and takes the default route (`rtm_dst_len == 0`) with the lowest metric.

Two things from your PR were missing in mine: logging the real error from `NLMSG_ERROR` instead of a stale `errno`, and retrying `recv()` after `EINTR`. I've added both to #3052 and credited this PR in the commit.

A couple of notes on this version, in case it's useful:

- The request is still `AF_UNSPEC`, and any main-table route with a gateway is accepted, not only the default one. When no IPv4 route in the main table has a gateway yet (for example before DHCP has finished, or on an IPv6-only network), the first match can be an IPv6 route. `inet_ntop(AF_INET, ...)` then turns the first 4 bytes of an address like `fe80::1` into `254.128.0.0`.
- About the 8584-byte datagram: on current kernels the size of each dump datagram follows the largest buffer the reader has passed to `recv()` so far (`max_recvmsg_len` in `net/netlink/af_netlink.c`, with `skb_reserve()` trimming the skb to that size). A reader that always uses 8 KiB gets datagrams of at most 8 KiB, so your standalone tool most likely used a larger buffer. That's why reading datagram by datagram with a fixed-size buffer is safe.

Either way, it's the maintainers' call which one to take. If they prefer yours, I'm happy to help test it.

